PR #323, opened 2014-06-21
Don't complain about undefined function

ghost:
When lazy loading UltiSnips on specific filetypes ([vim-plug](https://github.com/junegunn/vim-plug) and [NeoBundle](https://github.com/Shougo/neobundle.vim) can do that), UltiSnips#FileTypeChanged() function can be undefined while ftdetect/UltiSnips.vim is loading.


HerringtonDarkholme:
What about a `silent! call`?

PR #425, opened 2015-01-13
Fixes g:UltiSnipsEnableSnipMate "0" bechavior

ppalucki:
_vim.eval always returns string which evaluates to True in python. 

If value is set to "0" in vimrc, snipmate is used anyway.

It is not clear from documentation that this value have to be set to empty string "" to disable snipmate (I have lost couple of hours trying to disable snipmate).

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -95,7 +95,7 @@ class SnippetManager(object):
         enable_snipmate = True
         if _vim.eval("exists('g:UltiSnipsEnableSnipMate')") == "1":
             enable_snipmate = _vim.eval("g:UltiSnipsEnableSnipMate")
-        if enable_snipmate:
+        if enable_snipmate == "1":
             self.register_snippet_source("snipmate_files",
                 SnipMateFileSource())
 


pythonx/UltiSnips/snippet_manager.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, expand_trigger, forward_trigger, backward_trigger):
         enable_snipmate = True
         if _vim.eval("exists('g:UltiSnipsEnableSnipMate')") == "1":
             enable_snipmate = _vim.eval("g:UltiSnipsEnableSnipMate")
-        if enable_snipmate:
+        if enable_snipmate == "1":
```
SirVer: why not use `int(enable_snipmate)`, would that not handle strings and integers in the settings?

ppalucki: it was my first try, but it ends with error, when someone set "true"/"false"/'False"
and in unexpected behaviour when set to e.g. "3"   

IMHO it is also more coherent with docs, that explicitly says: only "1" value enable this feature

int(v) should be additionally backed by handling ValueError exception and that definitely is overhead to so simple functionality

ppalucki: frankly speaking I don't know, what is idiomatic way to this in vim (handling bool values in config), but from my simple experiments vim.eval always returns string no matter is it number or not - so expecting exactly == "1" in python will accept let x = 1 and x = "1"
and everything else will be treated as false 

from VIM docs about vim.eval 
returns ... "a string if the Vim expression evaluates to a string or number"

SirVer:
Merged. Thanks!

SirVer:
Rebasing seems to have confused github. The patch is in: https://github.com/SirVer/ultisnips/blob/master/pythonx/UltiSnips/snippet_manager.py#L98

PR #429, opened 2015-01-25
Allow package and module detection for the SnippetUtil.

s0undt3ch:


SirVer:
UltiSnips is language agnostic. What you added here is only useful for python. It should go into a `global !p` block in your configuration or if you want to contribute it, put it into https://github.com/honza/vim-snippets.

s0undt3ch:
Agreed. Thanks.

Issue #431, opened 2015-01-30
Langmap breaks python snippets


Dettorer:
I'm using the [bépo keymap](http://bepo.fr/wiki/Accueil) and I rebinded my vim with the following langmap:
````vim
set langmap=$`,\\"1,«2,»3,(4,)5,@6,+7,-8,/9,*0,=-,%=,bq,éw,pe,or,èt,^y,vu,di,lo,jp,z[,w],aa,us,id,ef,\\,g,ch,tj,sk,rl,n\\;,m',ç\\\\,ê<,àz,yx,xc,.v,kb,'n,qm,g\\,,h.,f/,#~,1!,2@,3#,4$,5%,6^,7&,8*,9(,0),°_,`+,BQ,ÉW,PE,OR,ÈT,!Y,VU,DI,LO,JP,Z{,W},AA,US,ID,EF,\\;G,CH,TJ,SK,RL,N:,M\\",Ç\\|,Ê>,ÀZ,YX,XC,:V,KB,?N,QM,G<,H>,F?
set langnoremap
````
Which works pretty well with plugins since [a recent vim update that included the `set langnoremap` option](https://groups.google.com/forum/#!msg/vim_dev/QnNwLWhJ744/1qNcD7d9OvgJ).

I also have a custom UltiSnips snippet for C and C++ header guards:
```python
# Guards
snippet once "Include header once only guard" !
#ifndef ${1:`!p
if not snip.c:
    name = re.sub(r'[^A-Za-z0-9]+','_', snip.fn).upper()
    snip.rv = ('%s' % name).upper()
else:
    snip.rv = snip.c`}
# define $1

${0}

#endif /* !$1 */
endsnippet
```
(My full vim configuration is also available [here](https://bitbucket.org/Dettorer/configuration/src))

Know if I enter the file (for example) `cpio.h` and insert "`once`" then expand the snippet, I get the following result:

Without using the langmap:

```C
#ifndef CPIO_H
# define CPIO_H



#endif /* !CPIO_H */
```
(Which is the expected behaviour)

And with the langmap (and with langnoremap activated):
```C
#ifndef CPIO_H
m+#%oJb~)/>
# define CPIO_H
m+#%oJb~)/>



#endif /* !CPIO_H
m+#%oJb~)/> */
```

Is it a bug in UltiSnips or is `set langnoremap` still not fully functional?

SirVer:
This is likely due to this hack: 

[...]

PR #444, opened 2015-02-12
Add support for vim's langnoremap new option

Dettorer:
This fixes the following issue:
https://github.com/SirVer/ultisnips/issues/431


SirVer:
Thanks! Merged.

PR #446, opened 2015-02-16
Context-aware snippets

seletskiy:
Hello again!

This PR is a proof-of-concept of snippets, that can be expanded not only by matching prefix/regexp, but by context.

Consider this example:

```
if true {
}<TAB>

// becomes

if true {
} else {
}

// but

type MyType struct {
}<TAB>

// becomes

type MyData struct {
}

func (data MyData) SomeMethod() {
}
```

So, no matter that same trigger is used (`}` in that case), different snippets are chosen for expansion.

I think it's very powerful mechanism, which will lead to just hitting `<TAB>` in most cases without typing at all.

It's not ready-to-go solution, and I'm glad to listen some thoughts.

In current implementation snippets use following syntax:

```
snippet } "Description" "is_if_bracket(b, l, c)" xw
} else {
	$1
}
endsnippet
```

Where `is_if_bracket` (note the `x` flag) is a python function that will return `True`, if cursor stands at closing bracket for `if` statement.

BTW, I've tried to fix syntax file for correct highlighting and found it nearly impossible to comprehend. Is there really solid reason to have triggers which can be defined using multiple nested double-quotes? As far as I understand it makes that patterns so complex.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/446)

### DIFF
--- a/pythonx/UltiSnips/snippet/definition/_base.py
+++ b/pythonx/UltiSnips/snippet/definition/_base.py
@@ -189,8 +189,8 @@ class SnippetDefinition(object):
 
         if match and self._context_code:
             self._context = self._context_match()
-            print(match, self._trigger, self._context)
-            match = self._context != None
+            if not self.context:
+                match = False
 
         return match
 
--- a/pythonx/UltiSnips/snippet/definition/_base.py
+++ b/pythonx/UltiSnips/snippet/definition/_base.py
@@ -4,6 +4,7 @@
 """Snippet representation after parsing."""
 
 import re
+
 import vim
 
 from UltiSnips import _vim
@@ -44,7 +45,7 @@ class SnippetDefinition(object):
     _TABS = re.compile(r"^\t*")
 
     def __init__(self, priority, trigger, value, description,
-                 options, globals, location, context=None):
+                 options, globals, location, context):
         self._priority = int(priority)
         self._trigger = as_unicode(trigger)
         self._value = as_unicode(value)
@@ -90,17 +91,17 @@ class SnippetDefinition(object):
         code = "\n".join([
             'import re, os, vim, string, random',
             '\n'.join(self._globals.get('!p', [])).replace('\r\n', '\n'),
-            'ctx["match"] = ' + self._context_code,
+            'context["match"] = ' + self._context_code,
             ''
         ])
 
         context = {'match': False}
         locals = {
-            'ctx': context,
-            'w': current.window,
-            'b': current.buffer,
-            'l': current.window.cursor[0],
-            'c': current.window.cursor[1],
+            'context': context,
+            'window': current.window,
+            'buffer': current.buffer,
+            'line': current.window.cursor[0],
+            'column': current.window.cursor[1],
         }
 
         exec(code, locals)
--- a/pythonx/UltiSnips/snippet/definition/_base.py
+++ b/pythonx/UltiSnips/snippet/definition/_base.py
@@ -104,9 +104,7 @@ class SnippetDefinition(object):
             'column': current.window.cursor[1],
             'cursor': current.window.cursor,
         }
-
         exec(code, locals)
-
         return context["match"]
 
     def has_option(self, opt):
@@ -142,7 +140,7 @@ class SnippetDefinition(object):
 
     @property
     def context(self):
-        """Returns matched context."""
+        """The matched context."""
         return self._context
 
     def matches(self, trigger):
--- a/pythonx/UltiSnips/snippet/definition/_base.py
+++ b/pythonx/UltiSnips/snippet/definition/_base.py
@@ -4,6 +4,7 @@
 """Snippet representation after parsing."""
 
 import re
+import vim
 
 from UltiSnips import _vim
 from UltiSnips.compatibility import as_unicode
@@ -43,7 +44,7 @@ class SnippetDefinition(object):
     _TABS = re.compile(r"^\t*")
 
     def __init__(self, priority, trigger, value, description,
-                 options, globals, location):
+                 options, globals, location, context):
         self._priority = int(priority)
         self._trigger = as_unicode(trigger)
         self._value = as_unicode(value)
@@ -53,6 +54,8 @@ class SnippetDefinition(object):
         self._last_re = None
         self._globals = globals
         self._location = location
+        self._context_code = context
+        self._context = None
 
         # Make sure that we actually match our trigger in case we are
         # immediately expanded.
@@ -77,6 +80,32 @@ class SnippetDefinition(object):
             self._last_re = match
             return match
         return False
+
+    def _context_match(self):
+        current = vim.current
+        # skip on empty buffer
+        if len(current.buffer) == 1 and current.buffer[0] == "":
+            return
+
+        code = "\n".join([
+            'import re, os, vim, string, random',
+            '\n'.join(self._globals.get('!p', [])).replace('\r\n', '\n'),
+            'ctx["match"] = ' + self._context_code,
+            ''
+        ])
+
+        context = {'match': False}
+        locals = {
+            'ctx': context,
+            'w': current.window,
+            'b': current.buffer,
+            'l': current.window.cursor[0],
+            'c': current.window.cursor[1],
+        }
+
+        exec(code, locals)
+
+        return context["match"]
 
     def has_option(self, opt):
         """Check if the named option is set."""
@@ -108,6 +137,11 @@ class SnippetDefinition(object):
     def location(self):
         """Where this snippet was defined."""
         return self._location
+
+    @property
+    def context(self):
+        """Returns matched context."""
+        return self._context
 
     def matches(self, trigger):
         """Returns True if this snippet matches 'trigger'."""
@@ -152,6 +186,12 @@ class SnippetDefinition(object):
             if text_before.strip(' \t') != '':
                 self._matched = ''
                 return False
+
+        if match and self._context_code:
+            self._context = self._context_match()
+            print(match, self._trigger, self._context)
+            match = self._context != None
+
         return match
 
     def could_match(self, trigger):
@@ -236,7 +276,8 @@ class SnippetDefinition(object):
 
         snippet_instance = SnippetInstance(
             self, parent, initial_text, start, end, visual_content,
-            last_re=self._last_re, globals=self._globals)
+            last_re=self._last_re, globals=self._globals,
+            context=self._context)
         self.instantiate(snippet_instance, initial_text, indent)
 
         snippet_instance.update_textobjects()
--- a/pythonx/UltiSnips/snippet/definition/_base.py
+++ b/pythonx/UltiSnips/snippet/definition/_base.py
@@ -44,7 +44,7 @@ class SnippetDefinition(object):
     _TABS = re.compile(r"^\t*")
 
     def __init__(self, priority, trigger, value, description,
-                 options, globals, location, context):
+                 options, globals, location, context=None):
         self._priority = int(priority)
         self._trigger = as_unicode(trigger)
         self._value = as_unicode(value)
--- a/pythonx/UltiSnips/snippet/definition/_base.py
+++ b/pythonx/UltiSnips/snippet/definition/_base.py
@@ -102,6 +102,7 @@ class SnippetDefinition(object):
             'buffer': current.buffer,
             'line': current.window.cursor[0],
             'column': current.window.cursor[1],
+            'cursor': current.window.cursor,
         }
 
         exec(code, locals)
--- a/pythonx/UltiSnips/snippet/source/file/ultisnips.py
+++ b/pythonx/UltiSnips/snippet/source/file/ultisnips.py
@@ -72,12 +72,11 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
             opts = words[-1]
             remain = remain[:-len(opts) - 1].rstrip()
 
-    if 'x' in opts:
+    context = None
+    if 'e' in opts:
         left = remain[:-1].rfind('"')
         if left != -1 and left != 0:
             context, remain = remain[left:].strip('"'), remain[:left]
-    else:
-        context = None
 
     # Get and strip description if it exists
     remain = remain.strip()
@@ -111,10 +110,12 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
     if snip == 'global':
         python_globals[trig].append(content)
     elif snip == 'snippet':
-        return 'snippet', (UltiSnipsSnippetDefinition(priority, trig, content,
-                                                      descr, opts, python_globals,
-                                                      '%s:%i' % (filename, start_line_index),
-                                                      context),)
+        definition = UltiSnipsSnippetDefinition(
+            priority, trig, content,
+            descr, opts, python_globals,
+            '%s:%i' % (filename, start_line_index),
+            context)
+        return 'snippet', (definition,)
     else:
         return 'error', ("Invalid snippet type: '%s'" % snip, lines.line_index)
 
--- a/pythonx/UltiSnips/snippet/source/file/ultisnips.py
+++ b/pythonx/UltiSnips/snippet/source/file/ultisnips.py
@@ -65,11 +65,20 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
     # Get and strip options if they exist
     remain = line[len(snip):].strip()
     words = remain.split()
+
     if len(words) > 2:
         # second to last word ends with a quote
         if '"' not in words[-1] and words[-2][-1] == '"':
             opts = words[-1]
             remain = remain[:-len(opts) - 1].rstrip()
+
+    if 'x' in opts:
+        left = remain[:-1].rfind('"')
+        if left != -1 and left != 0:
+            context, remain = remain[left:].strip('"'), remain[:left]
+    else:
+        context = None
+
 
     # Get and strip description if it exists
     remain = remain.strip()
@@ -105,7 +114,8 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
     elif snip == 'snippet':
         return 'snippet', (UltiSnipsSnippetDefinition(priority, trig, content,
                                                       descr, opts, python_globals,
-                                                      '%s:%i' % (filename, start_line_index)),)
+                                                      '%s:%i' % (filename, start_line_index),
+                                                      context),)
     else:
         return 'error', ("Invalid snippet type: '%s'" % snip, lines.line_index)
 
--- a/pythonx/UltiSnips/snippet/source/file/ultisnips.py
+++ b/pythonx/UltiSnips/snippet/source/file/ultisnips.py
@@ -79,7 +79,6 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
     else:
         context = None
 
-
     # Get and strip description if it exists
     remain = remain.strip()
     if len(remain.split()) > 1 and remain[-1] == '"':
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -203,18 +203,19 @@ class SnippetManager(object):
 
     @err_to_scratch_buffer
     def add_snippet(self, trigger, value, description,
-                    options, ft='all', priority=0):
+                    options, ft='all', priority=0, context=None):
         """Add a snippet to the list of known snippets of the given 'ft'."""
         self._added_snippets_source.add_snippet(ft,
                                                 UltiSnipsSnippetDefinition(priority, trigger, value,
-                                                                           description, options, {}, 'added'))
+                                                                           description, options, {}, 'added',
+                                                                           context))
 
     @err_to_scratch_buffer
-    def expand_anon(self, value, trigger='', description='', options=''):
+    def expand_anon(self, value, trigger='', description='', options='', context=None):
         """Expand an anonymous snippet right here."""
         before = _vim.buf.line_till_cursor
         snip = UltiSnipsSnippetDefinition(0, trigger, value, description,
-                                          options, {}, '')
+                                          options, {}, '', context)
 
         if not trigger or snip.matches(before):
             self._do_snippet(snip, before)
@@ -573,10 +574,11 @@ class SnippetManager(object):
         if not before:
             return False
         snippets = self._snips(before, False)
-        # prefer snippets with context if any
-        snippets_with_context = [s for s in snippets if s.context]
-        if snippets_with_context:
-            snippets = snippets_with_context
+        if snippets:
+            # prefer snippets with context if any
+            snippets_with_context = [s for s in snippets if s.context]
+            if snippets_with_context:
+                snippets = snippets_with_context
         if not snippets:
             # No snippet found
             return False
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -475,6 +475,7 @@ class SnippetManager(object):
         elif feedkey:
             _vim.command('return %s' % _vim.escape(feedkey))
 
+    @err_to_scratch_buffer
     def _snips(self, before, partial):
         """Returns all the snippets for the given text before the cursor.
 
@@ -572,6 +573,10 @@ class SnippetManager(object):
         if not before:
             return False
         snippets = self._snips(before, False)
+        # prefer snippets with context if any
+        snippets_with_context = [s for s in snippets if s.context]
+        if snippets_with_context:
+            snippets = snippets_with_context
         if not snippets:
             # No snippet found
             return False
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -476,7 +476,6 @@ class SnippetManager(object):
         elif feedkey:
             _vim.command('return %s' % _vim.escape(feedkey))
 
-    @err_to_scratch_buffer
     def _snips(self, before, partial):
         """Returns all the snippets for the given text before the cursor.
 
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -36,6 +36,7 @@ UltiSnips                                      *snippet* *snippets* *UltiSnips*
       4.7.1 Replacement String                  |UltiSnips-replacement-string|
       4.7.2 Demos                               |UltiSnips-demos|
    4.8 Clearing snippets                        |UltiSnips-clearing-snippets|
+   4.9 Context snippets                         |UltiSnips-context-snippets|
 5. UltiSnips and Other Plugins                  |UltiSnips-other-plugins|
    5.1 Existing Integrations                    |UltiSnips-integrations|
    5.2 Extending UltiSnips                      |UltiSnips-extending|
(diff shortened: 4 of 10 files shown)


SirVer:
That is a really cool idea. I like it a lot. It could replace the 'b' 'i' 'w' and 'r' option. I did not look at the implementation at all yet, I'd prefer to get the design of this right. So I immediately have a bunch of questions:

1) Where should the python functions be defined?
2) Should the snippet syntax definition be updated? I.e. I think about a version statement that defines what form the snippet definitions have? We could internally rewrite old snippet files into the new format. Or should we shoehorn this into the existing file format?
3) Should VimL functions also be supported? How?

I think a small design doc answering those questions would be nice. Would you be willing to come up with a Google docs? I found that to be the easiest form of quick collaboration on designs.

seletskiy:
@SirVer: ok, I've done with some document: https://docs.google.com/document/d/1L0AA1V1TCbdv6YeJKGkuPXpzIkSuCH3Drj7EQ-835bg

Hope it's clear enough.

SirVer:
Thanks! I hope that was not too much hassle. Could you make the document world-commentable or share it with SirVer@gmx.de so that I can add in questions and remarks?

In my mind, the document might grow into a suggestion for redoing the current snippet features using this one - I will add some propositions for that. But I think getting the feature in with the backwards compatible syntax as you suggest is probably the first and most reasonable thing to do.

seletskiy:
@SirVer: I've change shared mode, now document should be commentable.

SirVer:
I started adding comments and thoughts and restructured the document a bit. I still have some ideas of what I would like to add to it, but now I get access denied :(. Could you check the sharing settings again and go over my comments?

seletskiy:
@SirVer: I've accepted your edits and respond to your comments. Also, I've added you as commenter using e-mail you've specified, so, please, check, that you have access. It should work fine, I don't see any restrictions in the share options dialog.

seletskiy:
@SirVer: by the way, you can checkout lastest commit in this pull request, it's already works as described in document.

SirVer:
I will - I am traveling the next two weeks though, so I will likely not have a lot of time.

seletskiy:
@SirVer: Any news? I'm using it for a while now and I've satisfied how it works.

SirVer:
Sorry, real life keeps interfering :(. A definite design change that is needed is to not pass arguments to the function: I mean `is_if_bracket(b, l, c)` should be `is_if_bracket`. And the function should take kwargs, so that we can add new arguments in the future without killing backwards compatibility. 

what do you think?

seletskiy:
@SirVer: No. There is no backward compatibility issues. In the current design any valid python expression can be specified as condition, for example, this one:

`re.match('^\s+if err ', util.get_prev_nonempty_line(b, l))` (see my [go.snippets](https://github.com/seletskiy/dotfiles/blob/master/.vim/UltiSnips/go.snippets#L79)).

That way is more flexible than the way where you always should specify function accepting `kwargs`.

And keep in mind, that `b, l, c` stands only as shortcuts for `vim.current.window.buffer`, `vim.current.window.cursor[0]` and `vim.current.window.cursor[1]`.

That way allows to write and test very simple python functions, which has no knowledge about `vim` module at all, but can operate on given buffer and cursor position (like my `util.get_prev_nonempty_line` function). It's like a constructor for building complex conditions out of simple checks.

SirVer:
> That way is more flexible than the way where you always should specify function accepting kwargs.

It is not more flexible, it is just more terse. The other way of looking at this is: if we always require a function, a lambda can be used to use a one line expression. so there is nothing different requiring a function. 

> And keep in mind, that b, l, c stands only as shortcuts for vim.current.window.buffer, vim.current.window.cursor[0] and vim.current.window.cursor[1].

I think that is a design issue. Where are these abbreviations defined? In which scope do they reside? Where are they documented. Having on dictionary for the kwargs solves these issues.

I think if we want to give the users the full power of python, we should expect them to know how to use it, i.e. assume they understand the vim module. nothing against convenience variables, but I'd rather have them well defined.

seletskiy:
@SirVer: Honestly, I don't like that idea with `kwargs`.

With `kwargs` there are three ways of specifying contexts:

* Using `global` definition in snippet files:
    ```
    global !p
    def some_condition_function(buffer, line, **kwargs):
        return some_check_based_on_buffer_and_line(buffer, line)
    endglobal
    ```
    Snippet definition:
    ```
    snippet trigger "description" "some_condition_function" x
    ```

* Using function definitions somewhere in pythonx directory:
    ```
    def some_condition_function(buffer, line, **kwargs):
        return some_check_based_on_buffer_and_line(buffer, line)
    ```
    Snippet definition:
    ```
    snippet trigger "description" "some_module.some_condition_function" x
    ```

* Using `lambda`:
    Snippet definition:
    ```
    snippet trigger "description" "lambda buffer, line, **kwargs: some_check_based_on_buffer_and_line(buffer, line)" x
    ```

**Note,** that in every definition you need to write `**kwargs`, because of possible compatibility issues. The reason of this issues is that if function accept less arguments that are passed via `**kwargs`, the there will be exception.

And I have some library functions, which are not UltiSnips specific and have no knowledge about `**kwargs` stuff.

So either I have to add `**kwargs` to all my library functions, or either write a wrapper for every library function.

I see, that variant proposed by me is, maybe, more terse but laconic at the same path.

I totally agree, that "magic" variables `b, l, c` should be documented (I'll do it), and, I guess, should be renamed to `buffer, line, column` for clarity.

SirVer:
that is convincing me.

pythonx/UltiSnips/snippet/definition/_base.py
```diff
@@ -4,6 +4,7 @@
 """Snippet representation after parsing."""
 
 import re
+import vim
```
SirVer: please separate vim import from std library imports by one empty line

pythonx/UltiSnips/snippet/definition/_base.py
```diff
@@ -43,7 +44,7 @@ class SnippetDefinition(object):
     _TABS = re.compile(r"^\t*")
 
     def __init__(self, priority, trigger, value, description,
-                 options, globals, location):
+                 options, globals, location, context=None):
```
SirVer: please do not use default values, instead specific the context as None everywhere.

pythonx/UltiSnips/snippet/definition/_base.py
```diff
@@ -78,6 +81,32 @@ def _re_match(self, trigger):
             return match
         return False
 
+    def _context_match(self):
+        current = vim.current
+        # skip on empty buffer
+        if len(current.buffer) == 1 and current.buffer[0] == "":
```
SirVer: This requires some more tests. Also I do not think this is python3 compatible - I need to add that to the travis tests :/

pythonx/UltiSnips/snippet/definition/_base.py
```diff
@@ -78,6 +81,32 @@ def _re_match(self, trigger):
...
+        code = "\n".join([
+            'import re, os, vim, string, random',
+            '\n'.join(self._globals.get('!p', [])).replace('\r\n', '\n'),
+            'ctx["match"] = ' + self._context_code,
+            ''
+        ])
+
+        context = {'match': False}
+        locals = {
+            'ctx': context,
```
SirVer: please document and do not use single character names.

pythonx/UltiSnips/snippet/source/file/ultisnips.py
```diff
@@ -65,12 +65,20 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
...
     remain = line[len(snip):].strip()
     words = remain.split()
+
     if len(words) > 2:
         # second to last word ends with a quote
         if '"' not in words[-1] and words[-2][-1] == '"':
             opts = words[-1]
             remain = remain[:-len(opts) - 1].rstrip()
 
+    if 'x' in opts:
```
SirVer: Why did you choose 'x' as the option? Should this not be 'e' for expression? or 'p' for python?

pythonx/UltiSnips/snippet/source/file/ultisnips.py
```diff
@@ -65,12 +65,20 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
...
+        left = remain[:-1].rfind('"')
+        if left != -1 and left != 0:
+            context, remain = remain[left:].strip('"'), remain[:left]
+    else:
```
SirVer: remove the clause by initializing context to None before the if.

pythonx/UltiSnips/snippet/source/file/ultisnips.py
```diff
@@ -105,7 +113,8 @@ def _handle_snippet_or_global(filename, line, lines, python_globals, priority):
     elif snip == 'snippet':
         return 'snippet', (UltiSnipsSnippetDefinition(priority, trig, content,
```
SirVer: make a temporary for the UltiSnipsSnippetDefinition before the return, so that the construction becomes more readable.

pythonx/UltiSnips/snippet_manager.py
```diff
@@ -572,6 +574,11 @@ def _try_expand(self):
         if not before:
             return False
         snippets = self._snips(before, False)
+        if snippets:
```
SirVer: What is the reason to prefer snippets with contexts here?

seletskiy:
@SirVer: Thanks! I'll fix your notices soon.

In meantime, can you consider the following proposal about syntax highlighting of snippet files? There is overcomplicated patterns for highlighting snippet definition. The reason of that is right-to-left definition parsing. And the reason of that is quite unclear for me. As far as I can see from docs, it is possible to use any wrap character for quoting trigger text, but there is no practical reason for using this solution rather than just escaping double quote with `\`, which should make syntax highlighting match more simpler without considerable efforts of rewritting parser. Maybe I miss some point there, can you shed a light on that case?

pythonx/UltiSnips/snippet_manager.py
seletskiy: Because it is useful for having non-context snippets as default snippet with same trigger as context snippets.

Like this:
```
snippet i "default if clause" b
...
snippet i "if for error check" "is_err_declared_in_prev_statement()" bx
```

seletskiy:
@SirVer: I've fixed all marks.

SirVer:
Could you please merge master and ping me again? There are conflicts.

seletskiy:
@SirVer: fixed.

seletskiy:
@SirVer: ping? I have idea for next PR already.

SirVer:
What idea do you have?

This looks nearly good - however I miss tests. Could you add some for all the features?

---

Reviewed files:
- doc/UltiSnips.txt @ r3
- pythonx/UltiSnips/snippet/definition/_base.py @ r3
- pythonx/UltiSnips/snippet/definition/snipmate.py @ r2
- pythonx/UltiSnips/snippet/source/file/ultisnips.py @ r2
- pythonx/UltiSnips/text_objects/_snippet_instance.py @ r1
- syntax/snippets.vim @ r1
- test/test_UltiSnipFunc.py @ r4

---

<sup>**[doc/UltiSnips.txt, line 658 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZlosJsaSq14zknNho)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L658)):</sup>
80 chars limit exceeded? I am not sure, I am trying reviewable.io right now and it shows a weird line break here.

---

<sup>**[doc/UltiSnips.txt, line 827 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZlv0-LNWogaFAjfyQ)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L827)):</sup>
Indent the other entries for consistency so that the '-' are aligned again?

---

<sup>**[doc/UltiSnips.txt, line 1312 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZmDS5RszngPE2JQVd)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1312)):</sup>
The 'expression' ....

---

<sup>**[doc/UltiSnips.txt, line 1315 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZmG_hBe_ZzhSK6xYi)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1315)):</sup>
The following python modules are automatically imported into the scope before 'expression' is evaluated.

Also, the following variables are defined:

---

<sup>**[doc/UltiSnips.txt, line 1325 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZmfegRq9K5s3ysFip)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1325)):</sup>
...access the current line...

and drop the last word: 'expression'

---

<sup>**[doc/UltiSnips.txt, line 1333 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZmv6OjeLOnZ_VN39d)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1333)):</sup>
if the previous

---

<sup>**[doc/UltiSnips.txt, line 1335 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZmzCOWT1qH8xdE2-D)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1335)):</sup>
Also, will this not throw an error if the buffer is not long enough? It probably doesn't matter for an example, but having one that is not prone to crash would be nicer.

---

<sup>**[doc/UltiSnips.txt, line 1336 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZnBJADgcAmxjclV5P)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1336)):</sup>
I think that is clear from context that those will be preferred. Drop the paragraph?

---

<sup>**[doc/UltiSnips.txt, line 1356 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZnRlbA9OztJW6kHxX)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1356)):</sup>
the default 'if'

---

<sup>**[doc/UltiSnips.txt, line 1358 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZnTtmtc6OvKj0sqEl)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1358)):</sup>
It is a good idea... to a separate module

---

<sup>**[doc/UltiSnips.txt, line 1359 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZnnX2N0Rb1A1O9d1I)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1359)):</sup>
use module, not library - it is the term python uses. Also a few times below.

---

<sup>**[doc/UltiSnips.txt, line 1375 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZn_82_s5yXYmi9prR)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1375)):</sup>
if the cursor

---

<sup>**[doc/UltiSnips.txt, line 1376 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZnb6qhSiV_bNHktn1)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1376)):</sup>
depending on

---

<sup>**[doc/UltiSnips.txt, line 1380 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZnwJKsNgYOefu-DVt)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1380)):</sup>
this paragraph has weird line breaks 

---

<sup>**[doc/UltiSnips.txt, line 1381 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZo1OlYFY2yUvKCwy5)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1381)):</sup>
typo: considered

---

<sup>**[doc/UltiSnips.txt, line 1382 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZo4JW-qAh9-strZMb)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/doc/UltiSnips.txt#L1382)):</sup>
The evaluated value of 'condition' is available in the 'context' variable inside the snippet.

---

<sup>**[pythonx/UltiSnips/snippet/definition/_base.py, line 107 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZobkopMV1hduq5DR6)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/pythonx/UltiSnips/snippet/definition/_base.py#L107)):</sup>
remove empty line

---

<sup>**[pythonx/UltiSnips/snippet/definition/_base.py, line 109 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZod7vAqwjxZy_I6mQ)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/pythonx/UltiSnips/snippet/definition/_base.py#L109)):</sup>
remove empty line

---

<sup>**[pythonx/UltiSnips/snippet/definition/_base.py, line 145 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZofrtrKamZzBpU6WX)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/pythonx/UltiSnips/snippet/definition/_base.py#L145)):</sup>
"""The matched context."""

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 479 \[r4\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JnZpNm42r9PezjgMt1C)** ([raw file](https://github.com/sirver/ultisnips/blob/87c3a7ba43c6a80c6c150195a8cd1a8816a76d63/pythonx/UltiSnips/snippet_manager.py#L479)):</sup>
This is not okay. Other libraries unfortunately use this as API (for example YCM) and they expect to see exception from it. We cannot show them in a Vim window always. Please revert.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/446)

seletskiy:
> I think that is clear from context that those will be preferred. Drop the paragraph?
Nope, because it's new mechanism. Previously, when there are more than one snippet which are can be expanded on the same time, UltiSnips shows menu for selecting snippet.

In case of context + non-context snippet match, context snippet will be used unconditionally.

seletskiy:
> Also, will this not throw an error if the buffer is not long enough? It probably doesn't matter for an example, but having one that is not prone to crash would be nicer.

Nope, because in case of single-line buffer expression will be evaluated `buffer[line-2]` into `buffer[-1]`, which is equals to `buffer[0]`.

seletskiy:
@SirVer: fixed. Also, some basic tests added.

SirVer:
> In case of context + non-context snippet match, context snippet will be used unconditionally and non-context will be ignored.

I understood, but I think it is clear without this paragraph. Why else would one use context aware snippets if not to overwrite already existing ones?

---

Reviewed files:
- doc/UltiSnips.txt @ r5
- pythonx/UltiSnips/snippet/definition/_base.py @ r5
- pythonx/UltiSnips/snippet_manager.py @ r5
- test/test_ContextSnippets.py @ r5

---

<sup>**[test/test_ContextSnippets.py, line 87 \[r5\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JntGth61gaNKESp2R0g)** ([raw file](https://github.com/sirver/ultisnips/blob/904fbdecf54f265ce231f90cfea7284f2ff314c7/test/test_ContextSnippets.py#L87)):</sup>
What about

~~~
priority 1
snippet i "desc" "True" e
abc
endsnippet

priority 100
snippet i
once
endsnippet
~~~

I would expect the second snippet to be expanded always. Is that the case?

---

<sup>**[test/test_UltiSnipFunc.py, line 162 \[r5\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-JntHLuxLTDh9zdQhqbQ)** ([raw file](https://github.com/sirver/ultisnips/blob/904fbdecf54f265ce231f90cfea7284f2ff314c7/test/test_UltiSnipFunc.py#L162)):</sup>
Sorry, missed that before: 80 columns.

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/446)

seletskiy:
@SirVer: fixed. I've added test you proposed, works as expected.

Regarding documentation: I wanted to emphasize the possibility of making fallbacks with non-context snippets, because it's very useful. So, I reworded problematic paragraph to make it clear.

kovetskiy:
Hello @seletskiy and @SirVer, I'm trying to use this feature and I found big problem.

When I'm using some functions with some bugs, for example with syntax mistakes, or 'index out of range' errors, I can't hook this errors, my snippets just not work, when python expression has any errors.

I think it's wrong behaviour and @seletskiy should return back this line: https://github.com/seletskiy/ultisnips/commit/904fbdecf54f265ce231f90cfea7284f2ff314c7#diff-df9ffe28f13525264c92e4bbcbbc28c1L479 (line 479)

SirVer:
@kovetskiy Thanks for the feedback! This line must stay out of there for sure though - otherwise we break youcompleteme integration. But we should be able to surface the error on expanding. 

@seletskiy: how about adding a test with a syntax error in the condition and check that it throws an understandable error on expanding? That would catch the bug kovetskiy found. 

Otherwise one more nit in the documentation, the rest lgtm.

---

Reviewed files:
- doc/UltiSnips.txt @ r6
- test/test_ContextSnippets.py @ r6
- test/test_UltiSnipFunc.py @ r6

---

<sup>**[doc/UltiSnips.txt, line 1356 \[r6\]](https://reviewable.io:443/reviews/sirver/ultisnips/446#-Jny_OHipEaFAPW9Kble)** ([raw file](https://github.com/sirver/ultisnips/blob/1b3ecf4a85bf116e086869b2aa17e5cd118cbf0d/doc/UltiSnips.txt#L1356)):</sup>
It's a good idea to move context conditions into ....

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/446)

seletskiy:
@SirVer, @kovetskiy: it's not that easy. The reason I added `@err_to_scratch_buffer` before `_snips` call was exactly this issue. After digging around I've found, that `_snips` called in first place on vim startup. Checking traceback shows following:

```
  File "/home/s.seletskiy/.vim/bundle/YouCompleteMe/autoload/../python/ycm/youcompleteme.py", line 257, in OnBufferVisit
    _AddUltiSnipsDataIfNeeded( extra_data )
  File "/home/s.seletskiy/.vim/bundle/YouCompleteMe/autoload/../python/ycm/youcompleteme.py", line 395, in _AddUltiSnipsDataIfNeeded
    rawsnips = UltiSnips_Manager._snips( '', 1 )
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 488, in _snips
    print('\n'.join(traceback.format_stack()))
```

`_snips` call is not wrapped by `err_to_scratch_buffer`, and exception from invalid syntax goes directly to the YCM (which is completely ignorant to that exceptions and produces no output).
And it's looks like that subsequent calls of `_snips` will never try load nor buggy snippet, nor that follows (even if they are correct), because of `source.ensure()` call, which would cache loaded snippets util snippet file changed.

And I wonder, why YCM calls **private** `_snips` function in first place, breaking fundamental Python assumptions?

So, for clarification:

1. YCM triggered at vim startup;
2. YCM calls **private** `_snips` function;
3. `_snips` tries to load all defined snippets for filetype;
4. somewhere inside ultisnips all successfully parsed snippets got cached (`SnippetFileSource`, I guess);
5. `_snips` encounters buggy snippet and throws parsing exception; already loaded snippets left intact;
6. YCM catches exception and does nothing;
7. user triggers expansion of his buggy snippet;
8. ultisnips sees no reason to reload snippet file (which, in fact, will result in scratch buffer with error);
9. snippet is not expanded and user see no error produced.

Actually, this behaviour is already there and it happens with non-context snippets which are uses incorrect syntax. Just no output.

@SirVer: do you have any ideas how it should be fixed? I see only one way: rewrote `_snips` function i that way, so it will not cache loaded snippets on YCM invocation.

@SirVer: also, I've asked you several times but still did not get any answer regarding syntax highlighting issue of snippet files. Right now context syntax market with red background by syntax definition.

SirVer:
> And it's looks like that subsequent calls of _snips will never try load nor buggy snippet, nor that follows (even if they are correct), because of source.ensure() call, which would cache loaded snippets until snippet file will be changed.

It seems like that is our fault and that we can fix. 

> And I wonder, why YCM calls private _snips function in first place, breaking fundamental Python assumptions?

That is easy: YCM (and other plugins too) needed an interface to talk to UltiSnips. UltiSnips is not designed as a library, so there is no interface. They took what they could find, which was `_snips` - which is now by history a public API. The way forward would be to design a proper API for other plugins and deprecate `_snips`, but I do not have too much time - so UltiSnips is maintained, but not really further developed right now by me. 

> @SirVer: do you have any ideas how it should be fixed? I see only one way: rewrite _snips function in that way, so it will not cache loaded snippets on YCM invocation.

I think point 8 in your list is the flawed one - we should not try to cache a file with a known syntax error. 

> @SirVer: also, I've asked you several times but still did not get any answer regarding syntax highlighting issue of snippet files. Right now context syntax market with red background by syntax definition.

sorry, indeed. Feel free to change the syntax highlighting file, but use a separate PR, this one should go in ASAP. 

>  Is there really solid reason to have triggers which can be defined using multiple nested double-quotes?

Yes, multiword triggers require some form of quoting, but IIRC it does not need to be a `"`, so I think the rule could be simplified to be easier to understand and maybe not correct in all cases. That would be fine, since it would discourage people from actually writing hard to read triggers. 

fyi, I am on vacation for 10 days now and I will not check my email.

seletskiy:
@SirVer: okay, I found the way of gracefully fixing issue with error reporting. Look at latest commit, it will not "cache" file if it contains syntax errors.

SirVer:
Reviewed files:
- doc/UltiSnips.txt @ r7
- pythonx/UltiSnips/snippet/source/file/_base.py @ r8
- test/test_ContextSnippets.py @ r9

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/446)

SirVer:
Great, lgtm to me now. Thanks for this, Stanislav!

PR #456, opened 2015-03-05
remove duplicate things in vim_test_case.py

spacewander:


SirVer:
thanks for cleaning this up!

Issue #457, opened 2015-03-06
Cannot add text to consecutive tabstops


kawing-chiu:
Consider the following snippet:
```
snippet t "generic html tag" w
<$1`!p
snip.rv += ""
if t[2]:
  snip.rv += ' '
`$2>$3</$1>
endsnippet
```
When I type
```
t<Tab>img<C-j>src=
```
I am expecting to get
```
<img src=></img>
```
However, i get this:
```
<imgsrc=></imgsrc=>
```
That is, when `$2` is put "immediately" (in fact not, because content is added dynamically between `$1` and `$2`) after `$1`, I have no way to add text to `$2`. I consider this a bug because no matter how I arrange the tabstops I should be able to jump to and add text to it.

SirVer:
dupe of https://github.com/SirVer/ultisnips/issues/158

kawing-chiu:
Hi. Sorry for reporting a duplicate issue. After reading #158, it is not clear to me what the current status of this issue is. Will this be fixed in the foreseeable future?

SirVer:
An open bug means the problem is acknowledged and ready to be worked on. The constant problem is that there are more open bugs than active contributors (hint, hint). 

I have currently only enough time available to keep UltiSnips maintained - so I cannot commit on fixing it soon. If you are annoyed by it and want to fix it, I will provide guidance on how to go about it.

kawing-chiu:
Well, I have attempted to do something about it:
```
diff --git a/pythonx/UltiSnips/snippet_manager.py b/pythonx/UltiSnips/snippet_manager.py
index c12a529..06810d1 100644
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -319,7 +319,7 @@ class SnippetManager(object):
                     lt = '\n'.join(lt)
                     ct = '\n'.join(ct)
                     es = diff(lt, ct, initial_line)
-                self._csnippets[0].replay_user_edits(es)
+                self._csnippets[0].replay_user_edits(es, self._ctab)
             except IndexError:
                 # Rather do nothing than throwing an error. It will be correct
                 # most of the time
diff --git a/pythonx/UltiSnips/text_objects/_base.py b/pythonx/UltiSnips/text_objects/_base.py
index 24e9a21..6855b48 100644
--- a/pythonx/UltiSnips/text_objects/_base.py
+++ b/pythonx/UltiSnips/text_objects/_base.py
[...]

PR #462, opened 2015-03-13
Fix adjacent tabstop input issue.

kawing-chiu:
Previously, adjacent tabstops like $1$2 were not handled properly.
$2 could be jumped to but no text could be added. See issue #457.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1444,6 +1444,7 @@ individuals have contributed to UltiSnips (in chronological order):
    Stanislav Seletskiy - seletskiy
    Pawel Palucki - ppalucki
    Dettorer - dettorer
+   Zhao Jiarong - kawing-chiu
 
 
 Thank you for your support.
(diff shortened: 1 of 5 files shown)


SirVer:
This looks awesome, a very high quality contribution even with tests! Thanks for working on this. Could you please add yourself to the list of contributors in doc/UltiSnips and then this is ready for merge.

kawing-chiu:
Sorry for the long delay. Have been busy for quite a while. Ultisnips is a project of high quality, so I'm just trying to keep up the standard :)

While working on the fix, I've found that this fix only works for simple situaions but not complex one like:
```
snippet test ""
hi ${1:this ${2:second ${3}}}$4
endsnippet
```
In the above snippet `$4` cannot be jumped and added text to. I tried but failed to find a similar "duck typing" check solution for it. It seems to me that the `_do_edit` function is becoming a monolithic and forbidding monster that is hard to deal with.

SirVer:
> It seems to me that the _do_edit function is becoming a monolithic and forbidding monster that is hard to deal with. 

I fear you are right. 

Even if this does not fix all issues, it is certainly an improvement. Gonna merge. Thanks for the contibution.

PR #480, opened 2015-04-12
Delete neocomplete.sources default setting

KazuakiM:
Hi,

I think "min_pattern_length", "max_candidates" and "is_volatile" are controlled by user setting.
So they are deleted.

### DIFF
--- a/autoload/neocomplete/sources/ultisnips.vim
+++ b/autoload/neocomplete/sources/ultisnips.vim
@@ -9,9 +9,6 @@ let s:source = {
    \ 'matchers' :
       \ (g:neocomplete#enable_fuzzy_completion ?
       \ ['matcher_fuzzy'] : ['matcher_head']),
-   \ 'min_pattern_length' : 1,
-   \ 'max_candidates' : 20,
-   \ 'is_volatile': 1,
    \ }
 
 function! s:source.gather_candidates(context)


SirVer:
@Shougo Can you comment on this? I do not know neocomplete.

Shougo:
Yes, it is just default value.
I think "max_candidates" can be deleted.
But I don't know if "min_pattern_length" and "is_volatile" can be deleted.

KazuakiM:
I think completion's word length is designed by user. I want to use completion's word length "3" and
 the same level's other candidates.

My setting.
```vim
" the default setting "2" by neocomplete.vim .
let g:neocomplete#auto_completion_start_length = 3
```
And I think "is_volatile" setting is 0(default) faster than 1.

SirVer:
> And I think "is_volatile" setting is 0(default) faster than 1.

of course it's faster, but it will also be wrong :). If the snippet files change you will not get updated snippets. 

> But I don't know if "min_pattern_length" and "is_volatile" can be deleted.

I still do not know what to make of this pull request, @Shougo Shougo. Who else would know?

Shougo:
> of course it's faster, but it will also be wrong :). If the snippet files change you will not get updated snippets. 

neocomplete may have that problem.  I will fix it later.

> I still do not know what to make of this pull request, @Shougo Shougo. Who else would know?

I think UltiSnips should not overwrite "min_pattern_length".
neosnippet source does not overwrite it in same reason.

Shougo:
@SirVer I checked neocomplete code.

> of course it's faster, but it will also be wrong :). If the snippet files change you will not get updated snippets. 

The cache is cleared in InsertLeave.  So "is_volatile" is not needed.

SirVer:
Okay, sounds like the change is correct as is. Thanks, gonna merge!

KazuakiM:
Thank you too :)

PR #482, opened 2015-04-14
Verify python support for UltiSnipsUsePythonVersion

blueyed:


autoload/UltiSnips/bootstrap.vim
```diff
@@ -20,10 +22,18 @@ function! UltiSnips#bootstrap#Bootstrap()
        endif
        let g:UltiSnipsUsePythonVersion = "<tab>"
    else
-       if g:UltiSnipsUsePythonVersion == 2
-           let g:_uspy=":py "
-       else
-           let g:_uspy=":py3 "
+       " Use user-provided value, but check if it's available.
```
SirVer: I do not understand this comment. Could you clarify?

SirVer:
Good change. Could you clarify just the one comment?

autoload/UltiSnips/bootstrap.vim
blueyed: It uses `g:UltiSnipsUsePythonVersion`, but still checks if it's available in Vim.
Before this, you could set it to `3`, but `has('python3')` might have been false and then followup errors were happening.

SirVer: Sorry, my comment was not clear: I meant to expand on the comment in the code so that the next reader has the context we already have from the pull request.

SirVer:
Could you also please merge/rebase master?

blueyed:
Done.

SirVer:
Thanks!

PR #486, opened 2015-04-25
Add more precise undo levels (expanding, jumping)

seletskiy:
Current behaviour is very annoying, when you want to undo snippet expansion and undo grabs a piece of your changes along with snippet.

Breaking undo achieved through re-setting &undolevel:

> Setting the value of 'undolevels' also breaks undo. Even when the new value is equal to the old value.
    [:h :undoj]

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/486)

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -111,6 +111,7 @@ class SnippetManager(object):
     def jump_forwards(self):
         """Jumps to the next tabstop."""
         _vim.command('let g:ulti_jump_forwards_res = 1')
+        _vim.command('let &undolevels = &undolevels')
         if not self._jump():
             _vim.command('let g:ulti_jump_forwards_res = 0')
             return self._handle_failure(self.forward_trigger)
@@ -119,6 +120,7 @@ class SnippetManager(object):
     def jump_backwards(self):
         """Jumps to the previous tabstop."""
         _vim.command('let g:ulti_jump_backwards_res = 1')
+        _vim.command('let &undolevels = &undolevels')
         if not self._jump(True):
             _vim.command('let g:ulti_jump_backwards_res = 0')
             return self._handle_failure(self.backward_trigger)
@@ -575,13 +577,15 @@ class SnippetManager(object):
         if not snippets:
             # No snippet found
             return False
-        elif len(snippets) == 1:
+        _vim.command('let &undolevels = &undolevels')
+        if len(snippets) == 1:
             snippet = snippets[0]
         else:
             snippet = _ask_snippets(snippets)
             if not snippet:
                 return True
         self._do_snippet(snippet, before)
+        _vim.command('let &undolevels = &undolevels')
         return True
 
     @property
--- a/test/test_Editing.py
+++ b/test/test_Editing.py
@@ -6,8 +6,8 @@ from test.constant import *
 
 class Undo_RemoveMultilineSnippet(_VimTest):
     snippets = ('test', 'Hello\naaa ${1} bbb\nWorld')
-    keys = 'test' + EX + ESC + 'u' + 'inothing'
-    wanted = 'nothing'
+    keys = 'test' + EX + ESC + 'u'
+    wanted = 'test'
 
 
 class Undo_RemoveEditInTabstop(_VimTest):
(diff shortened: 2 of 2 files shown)


seletskiy:
@SirVer: I'll fix failed tests (they are obviously linked with undo). Tell me what do you think about it?

SirVer:
I was playing around with what `&undolevels` does, but I do not get it. Is setting undolevels to itself basically the same as exiting insert mode and reentering it? If so, I think this is a reasonable change. 

---

Reviewed files:

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 117 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/486#-JntJ9w-lCSw7pBMrnbR)** ([raw file](https://github.com/sirver/ultisnips/blob/e920a8bd8dfc0814418c70002c0675c6222c22af/pythonx/UltiSnips/snippet_manager.py#L117)):</sup>
I am not sure if this should not only happen on success for everything below too?

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 573 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/486#-JntJMkjx6R9wd9zoS1v)** ([raw file](https://github.com/sirver/ultisnips/blob/e920a8bd8dfc0814418c70002c0675c6222c22af/pythonx/UltiSnips/snippet_manager.py#L573)):</sup>
This for example seems awkward: you tried to expand, but there was nothing to expand - still, your undo is now broken. Is that the behavior you had in mind?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/486)

seletskiy:
@SirVer: actually, I don't seem much difference when I reordering 'let &undolevel' statement. Looks like vim/documentation bug, but on my cases undo breaks as expected. Anyway, I'll place statements in more appropriate lines just in case. I'll ping you when tests will be fixed.

seletskiy:
@SirVer: ping. I've fixed failed tests and add some new tests to show implemented behaviour. Looks sane for me.

SirVer:
still confused about the undolevels for jumping. I think if we cannot jump, we should not touch them. 

---

Reviewed files:
- pythonx/UltiSnips/snippet_manager.py @ r2
- test/test_Editing.py @ r2

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 114 \[r2\]](https://reviewable.io:443/reviews/sirver/ultisnips/486#-JnyV-t05wTpO2xyMjRr)** ([raw file](https://github.com/sirver/ultisnips/blob/099d3bcfbc6bbbf738a6ede4c4d7b0c5ffad6eec/pythonx/UltiSnips/snippet_manager.py#L114)):</sup>
same here: only do that if we actually successfully jumped? i.e. if self._jump() returns True

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 123 \[r2\]](https://reviewable.io:443/reviews/sirver/ultisnips/486#-JnyV65uIsJ0kDYWoPwx)** ([raw file](https://github.com/sirver/ultisnips/blob/099d3bcfbc6bbbf738a6ede4c4d7b0c5ffad6eec/pythonx/UltiSnips/snippet_manager.py#L123)):</sup>
same here

---

<sup>**[test/test_Editing.py, line 35 \[r2\]](https://reviewable.io:443/reviews/sirver/ultisnips/486#-JnyVX2itQ-7rw6dPD6B)** ([raw file](https://github.com/sirver/ultisnips/blob/099d3bcfbc6bbbf738a6ede4c4d7b0c5ffad6eec/test/test_Editing.py#L35)):</sup>
isn't this unexpected? How can twice undo give the same result as once undo? Maybe increase undolevel in the vim config used for these tests?

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/486)

seletskiy:
> still confused about the undolevels for jumping. I think if we cannot jump, we should not touch them. 

The reason undo breaks before jump is that if I will undo jump, I expect to see my cursor where it was located before jump, not after. And actually I can't imagine real problem with undo breaking after trying to jump when there is nowhere to jump.

> isn't this unexpected? How can twice undo give the same result as once undo? Maybe increase undolevel in the vim config used for these tests?

It's not the same result. Please, look carefully, there is **two** expansions if `i` snippet, so first undo will revert expansion (and leave entered `i`), and the second undo will revert entering of `i`:

```
i|         <Tab>

if:
    |      i

if:
    i|     <Tab>

if:
    if:
        |  u

if:
    i|     u

if:
    |
```

SirVer:
okay, thanks for explaining.

---









Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/486)

PR #506, opened 2015-05-27
`augroup END` in ftdetect/UltiSnips.vim breaks other ftdetect scripts

dyng:
I had a trouble with [kchmck/vim-coffee-script](https://github.com/kchmck/vim-coffee-script) that filetype detection auto commands are not defined within 'filetypedetect' augroup. After hours of debugging, I found it's a problem with UltiSnippet (and some other plugins).

From the [user guide](https://github.com/vim/vim/blob/master/runtime/doc/filetype.txt#L177-L185), autocommand in ftdetect scripts should not be surrounded with an augroup block.

During the startup of vim, ftdetect scripts are executed at the below line in `filetype.vim`.

https://github.com/vim/vim/blob/master/runtime/filetype.vim#L2713

You can find that ftdetect scripts are executed in 'filetypedetect' augroup, but `augroup END` can enclose 'filetypedetect' group unexpectedly, makes following auto commands (of plugins besides `UltiSnips`) be defined in default autogroup, but not 'filetypedetect'.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/506)

### DIFF
--- a/ftdetect/UltiSnips.vim
+++ b/ftdetect/UltiSnips.vim
@@ -7,8 +7,5 @@ if exists("vimpager")
 endif
 
 if has("autocmd")
-   augroup UltiSnipsFileType
-      au!
-      autocmd FileType * call UltiSnips#FileTypeChanged()
-   augroup END
+  autocmd FileType * call UltiSnips#FileTypeChanged()
 endif
--- a/ftdetect/UltiSnips.vim
+++ b/ftdetect/UltiSnips.vim
@@ -7,5 +7,11 @@ if exists("vimpager")
 endif
 
 if has("autocmd")
-  autocmd FileType * call UltiSnips#FileTypeChanged()
+    augroup UltiSnipsFileType
+        au!
+        autocmd FileType * call UltiSnips#FileTypeChanged()
+    augroup END
+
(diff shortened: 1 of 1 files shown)


SirVer:
@dyng Sorry for the inconvenience and thanks for tracking this down - it is so easy to mess things up with Vim, it is just so brittle :/.

I think instead of using an augroup, we can also delete each individuall set autogroup here. Then we do not need an augroup at all. Have you tried that?

dyng:
Um.. sorry about the fact that I'm not familiar with this project and do not fully know what is this autogroup originally coded for. If this autogroup is defined for later deleting by group, I don't think this is a bad idea, maybe putting these code in a ftdetect script looks a bit strange though.

My opinion, defining Ultisnips autocmd in his own augroup then restoring filetypedetect group is an acceptable solution. But if you find there is a more elegant solution, feel free to change my code.

SirVer:
Okay, let's go with this for now then. Thanks for looking into this!

I added your name to the end of the doc/UltiSnips.txt contributor list. Could you verify that you are fine with the way you appear there?

dyng:
Thank you, I'm glad to be there. :)

PR #514, opened 2015-06-11
[RFC] Fire autocommands when setting up and tearing down inner mappings

wincent:
It can be useful to set up mappings that apply only during expansion of
a snippet. UltiSnips does this internally with the `_map_inner_keys` and
`_unmap_inner_keys` methods.

This commit adds some `User` autocommands so that users of UltiSnips can
set up their own mappings and tear them down at the same time as
`_map_inner_keys` and `_unmap_inner_keys`.

This is particularly useful for people who are writing their own
expansion and jump functions using `UltiSnips#JumpForwards` and
`UltiSnips#ExpandSnippet()` and friends. Here's an example from my own
dotfiles:

- https://github.com/wincent/wincent/commit/0664b627e7390c7ab5b047ee5818801481d0f46c
- https://github.com/wincent/wincent/commit/3740c248ee498220f3e7361b5e41023bd6cb2dbf

Using this approach I've been able to get rid of Supertab and have a
more nuanced autocompletion experience that works exactly as I'd like
with YouCompleteMe.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/514)

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -12,7 +12,8 @@ UltiSnips                                      *snippet* *snippets* *UltiSnips*
    3.1 Commands                                 |UltiSnips-commands|
    3.2 Triggers                                 |UltiSnips-triggers|
       3.2.1 Using your own trigger functions    |UltiSnips-trigger-functions|
-      3.2.2 Path to Python Module               |UltiSnips-python-module-path|
+      3.2.2 Overriding inner mappings           |UltiSnips-inner-mappings|
+      3.2.3 Path to Python Module               |UltiSnips-python-module-path|
    3.3 Snippet Search Path                      |UltiSnips-snippet-search-path|
    3.4 Warning About Select Mode Mappings       |UltiSnips-warning-smappings|
    3.5 Functions                                |UltiSnips-functions|
@@ -302,7 +303,33 @@ then you define your mapping as >
 and if the you can't expand or jump from the current location then the
 alternative function IMAP_Jumpfunc('', 0) is called.
 
-3.2.2 Path to Python module                   *UltiSnips-python-module-path*
+3.2.2 Inner mappings                          *UltiSnips-inner-mappings*
+--------------------
+
+                                *UltiSnipsMapInnerKeys* *UltiSnipsUnmapInnerKeys*
+For maximum compatibility with other plug-ins, UltiSnips sets up some special
+mappings when a snippet starts being expanded, and tears them down once
+expansion is completed. In order to make it possible to override these inner
+mappings, it fires the following "User" autocommands:
+
+UltiSnipsMapInnerKeys
+UltiSnipsUnmapInnerKeys
+
+For example, to call a pair of custom functions in response to these events,
+you might do: >
+
+   autocmd! User UltiSnipsMapInnerKeys
+   autocmd User UltiSnipsMapInnerKeys call CustomInnerKeyMapper()
+   autocmd! User UltiSnipsUnmapInnerKeys
+   autocmd User UltiSnipsUnmapInnerKeys call CustomInnerKeyUnmapper()
+
+Note that because snippet expansion may be nested, it is possible that
+|UltiSnipsMapInnerKeys| may be fired multiple times (as you enter nested
+snippets), but |UltiSnipsUnmapInnerKeys| will only fire once all nested
+snippets have been exited.
+
+
+3.2.3 Path to Python module                   *UltiSnips-python-module-path*
 ---------------------------
 
 For even more advanced usage, you can directly write python functions using
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -344,6 +344,7 @@ class SnippetManager(object):
                      ' <C-R>=UltiSnips#JumpBackwards()<cr>')
         _vim.command('snoremap <buffer> <silent> ' + self.backward_trigger +
                      ' <Esc>:call UltiSnips#JumpBackwards()<cr>')
+        _vim.command('silent doautocmd User UltiSnipsMapInnerKeys')
         self._inner_mappings_in_place = True
 
     def _unmap_inner_keys(self):
@@ -351,7 +352,7 @@ class SnippetManager(object):
         if not self._inner_mappings_in_place:
             return
         try:
-            _vim.command('silent doautocmd User UltiSnipsSnippetDone')
+            _vim.command('silent doautocmd User UltiSnipsUnmapInnerKeys')
             if self.expand_trigger != self.forward_trigger:
                 _vim.command('iunmap <buffer> %s' % self.forward_trigger)
                 _vim.command('sunmap <buffer> %s' % self.forward_trigger)
@@ -532,7 +533,6 @@ class SnippetManager(object):
         """Expands the given snippet, and handles everything that needs to be
         done with it."""
         self._map_inner_keys()
-        _vim.command('silent doautocmd User UltiSnipsDoSnippet')
 
         # Adjust before, maybe the trigger is not the complete word
         text_before = before
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -351,6 +351,7 @@ class SnippetManager(object):
         if not self._inner_mappings_in_place:
             return
         try:
+            _vim.command('silent doautocmd User UltiSnipsSnippetDone')
             if self.expand_trigger != self.forward_trigger:
                 _vim.command('iunmap <buffer> %s' % self.forward_trigger)
                 _vim.command('sunmap <buffer> %s' % self.forward_trigger)
@@ -531,6 +532,7 @@ class SnippetManager(object):
         """Expands the given snippet, and handles everything that needs to be
         done with it."""
         self._map_inner_keys()
+        _vim.command('silent doautocmd User UltiSnipsDoSnippet')
 
         # Adjust before, maybe the trigger is not the complete word
         text_before = before
--- a/test/test_Autocommands.py
+++ b/test/test_Autocommands.py
@@ -1,0 +1,31 @@
+# encoding: utf-8
+from test.vim_test_case import VimTestCase as _VimTest
+from test.constant import *
+
+# Autocommands  {{{#
+
+class Autocommands(_VimTest):
+    snippets = ('test', '[ ${1:foo} ]')
+    args = ''
+    keys = 'test' + EX + JF + ESC + \
+            ':execute "normal iM" . g:mapper_call_count . "\<Esc>"' + "\n" + \
+            ':execute "normal aU" . g:unmapper_call_count . "\<Esc>"' + "\n"
+    wanted = '[ foo M1U1]'
+
+    def _extra_options_pre_init(self, vim_config):
+        vim_config.append('let g:mapper_call_count = 0')
+        vim_config.append('function! CustomMapper()')
+        vim_config.append('  let g:mapper_call_count += 1')
+        vim_config.append('endfunction')
+
+        vim_config.append('let g:unmapper_call_count = 0')
+        vim_config.append('function! CustomUnmapper()')
+        vim_config.append('  let g:unmapper_call_count += 1')
+        vim_config.append('endfunction')
+
+        vim_config.append('autocmd! User UltiSnipsMapInnerKeys')
+        vim_config.append('autocmd User UltiSnipsMapInnerKeys call CustomMapper()')
+        vim_config.append('autocmd! User UltiSnipsUnmapInnerKeys')
+        vim_config.append('autocmd User UltiSnipsUnmapInnerKeys call CustomUnmapper()')
+
+# end: Autocommands  #}}}


wincent:
Labeling this as an RFC because the names of the autocommands probably suck, and we may want to tweak exactly where/when they fire. Also, may not even be a great idea to do this.

Let me know if you'd like to hear any more details about the motivating use case.

wincent:
So I thought I'd give a little context about the motivating use case here.

I basically want to use `<Tab>` as my principal key for interacting with UltiSnips and YouCompleteMe. Up until now I've been using Supertab to make this happen, but the combination has some glitches. Specifically, if you want `<Tab>` to mean both "accept/next completion" (for YouCompleteMe) and "jump to next placeholder" (for UltiSnips), which one should take precedence when you're simultaneously inside a placeholder and looking at a YouCompleteMe completion menu?

Up until recently, [the set-up I was using](https://github.com/wincent/wincent/blob/e1e071ee606ab6b6fbfb9bf6100464ed031eacfd/roles/dotfiles/files/.vim/plugin/autocomplete.vim) meant that UltiSnips' "jump to next placeholder" was always winning, leading to poor user experience. The example I linked to in the initial PR text is this:

> I have a snippet `log` that expands to `console.log();`, with a placeholder inside the parens. Now, say I have a file with:
>
>     var fooBar = 1;
>
> and I type:
>
>     log<tab>
>
> UltiSnips puts me here:
>
>     console.log(|);
>                 ^
>                 (cursor here)
>
> If I then type:
>
>     console.log(fooB<tab>);
>
> I wind up here:
>
>     console.log(fooB);|
>                       ^
>                       (cursor here)
>
> When what I really want is:
>
>     console.log(fooBar|);
>                       ^
>                       (cursor here)

All of the solutions I've seen for this so far have this flaw (ie. making UltiSnips take priority over YouCompleteMe in the above scenario), including the solutions on [this Stack Overflow page](http://stackoverflow.com/questions/14896327/ultisnips-and-youcompleteme) (which is currently the top Google hit for "ultisnips youcompleteme").

So, the motivation for the PR is to make it possible to swap this priority ordering, such that YouCompleteMe takes priority inside placeholders of expanded snippets. This is desirable because intuitively, when the autocomplete menu is visible (as it automatically is with YouCompleteMe when completion is available), you would intuitively expect `<Tab>` to interact with it, instead of with the invisible and implicit next placeholder managed by UltiSnips.

To make this work, I have to do the following:

* [Configure UltiSnips to use `<Tab>` and `<S-Tab>`](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/plugin/autocomplete.vim#L2-L4)
* [Configure YouComplete to *not* use `<Tab>` and `<S-Tab>`](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/plugin/autocomplete.vim#L13-L14)
* [Make use of the autocommand implemented in this PR to call a function which will override the "inner" mappings that UltiSnips will set up](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/plugin/autocomplete.vim#L9-L11)
* [Make sure that UltiSnips doesn't blow away our custom `SELECT` mode mappings](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/plugin/autocomplete.vim#L6-L7)
* [Set up mappings that overwrite UltiSnips `<Tab>` and `<S-Tab>` mappings to call my custom autocomplete function](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/after/plugin/UltiSnips.vim#L7-L14)
* [Use `let did_UltiSnips_after=1` to prevent UltiSnips from overwriting our custom mappings, which it will do in an effort to override Supertab, which I've actually uninstalled because it is no longer required by my new set-up](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/after/plugin/UltiSnips.vim#L1-L3)
* [In the autocommand callback, set up my own "inner" mappings](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/autoload/autocomplete.vim#L4-L12)
* [Additionally, set up a `<CR>` mapping to provide us with an unambiguous way of accepting a YouCompleteMe autocompletion inside an UltiSnips placeholder](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/autoload/autocomplete.vim#L17-L18) (while `<Tab>` has the standard "next completion" behavior)
* [In the other autocommand callback, tear down my own "inner" mappings](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/autoload/autocomplete.vim#L24-L25)
* [Finally, implement the custom function that implements our special autocomplete-or-expand/jump logic](https://github.com/wincent/wincent/blob/9a04973d2a4e886ac6560de8894d24b6054b08a0/roles/dotfiles/files/.vim/autoload/autocomplete.vim#L34-L67)

A couple of observations about the approach:

* I initially wanted to configure UltiSnips to not map any keys at all (since my sole objective is to entirely override them), but there are multiple places throughout the codebase that assume that a non-empty mapping is in place.
* I therefore considered just mapping the unwanted UltiSnips actions to useless/unreachable keys (like `<F18>`, `<F19>` etc), but without the autocommands added by the PR, I'd end up having to set up global maps rather than temporary, buffer-specific "inner" maps like UltiSnips does; keeping the maps narrowly scoped seems like a desirable property.

So, that's the motivation and the explanation of how it all fits together. With this, `<Tab>` can be comfortably used to mean "expand snippet", "next placeholder" and "next completion". `<CR>` serves as the "tie-breaker" that enables you to get out of completion mode when you're inside a placeholder and really just want to move to the next placeholder. I've been using this heavily for a couple of days now and it is awesome.

Let me know if you have any questions or comments, and thanks for making such a robust, extensible plug-in.

pythonx/UltiSnips/snippet_manager.py
```diff
@@ -531,6 +532,7 @@ def _do_snippet(self, snippet, before):
         """Expands the given snippet, and handles everything that needs to be
         done with it."""
         self._map_inner_keys()
+        _vim.command('silent doautocmd User UltiSnipsDoSnippet')
```
wincent: I was thinking that this perhaps should be moved inside `_map_inner_keys()` for parity with what we do in `_unmap_inner_keys()`, where the autocommand is fired from inside the method.

Additionally, I was thinking that the autocommand names themselves might be a bit too broad. We could alternatively consider naming them something more specifically as `UltiSnipsMapInnerKeys` and `UltiSnipsUnmapInnerKeys`.

(Naming is hard.)

lencioni:
I have been trying to use UltiSnips along with YouCompleteMe in exactly the same way as you describe here, and have suffered from what sounds like exactly the same pain points. I don't have any suggestions on naming or placement, but work that allows me to get things set up in a better way sounds really great to me, so thank you for spending time on this and sharing your motivations and approach! It is great to see how such a small and simple change to this plugin can extend its possibilities this much.

The only constructive comment I have is that it would be really great to add documentation for this so that others can benefit from the feature. But, I assumed that you would do that once @SirVer gives you the green light.

Let me know if there's anything I can do to help get this merged--it would certainly help to smooth out some of the rough edges I've been experiencing.

SirVer:
I think the approach is valid and having user autocommands seems sensible to me too. My only concern is that users are tempeted to do other stuff than just setting up mappings - for example modify the buffer in some way in these autocommands. That will then kill snippet expansion. 

Could you try adding a test and documentation for this change?

SirVer:
Review status: all files reviewed, 2 unresolved discussions, all commit checks successful.
Reviewed files:

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 354 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/514#-JrprTAUSSXBN9mov90S)** ([raw file](https://github.com/sirver/ultisnips/blob/88cf17c39232c50a86896089f210261acc71fdf8/pythonx/UltiSnips/snippet_manager.py#L354)):</sup>
The name is not ideal, because this autocommand will fire when all snippets are done. I.e. if you have 2 nested snippets running, it will only fire once you have exited both of them. Also to discourage doing anything else but modifying maps, I suggest calling it UltiSnipsUnmapInnerKeys. The other one can than be called UltiSnipsMapInnerKey for symmetry.

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 535 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/514#-JrWSZRZM9W6nTKG5FYu-r1-535)** ([raw file](https://github.com/sirver/ultisnips/blob/88cf17c39232c50a86896089f210261acc71fdf8/pythonx/UltiSnips/snippet_manager.py#L535)):</sup>
yes, please move into the _map_inner_keys() method.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/514)

wincent:
Wow, I just looked at the tmux-based test harness you have set up... That is seriously brilliant; I think I'm going to steal the idea for my own plug-in.

SirVer:
The test setup is responsible for the high stability of UltiSnips - so I agree that it is awesome. It predates the idea of testing vim plugins though. There are some good alternatives that I suggest looking into before stealing the one in UltiSnips. Some colleagues of mine wrote https://github.com/google/vroom - which cannot quite do interactive testing as we do, but is close.

Our test setup suffers from idiosyncrasies of encoding issues in python, vim and tmux and how they interact (poorly), so it is occasionally finicky.

SirVer:
@wincent any progress?

wincent:
I think it's done at this point (has tests, docs, and I've been running this code non-stop).

SirVer:
Could you rebase on master?

test/test_Autocommands.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Autocommands(_VimTest):
+    snippets = ('test', '[ ${1:foo} ]')
+    args = ''
+    keys = 'test' + EX + JF + ESC + \
+            ':execute "normal iM" . g:mapper_call_count . "\<Esc>"' + "\n" + \
+            ':execute "normal aU" . g:unmapper_call_count . "\<Esc>"' + "\n"
+    wanted = '[ foo M1U1]'
+
+    def _extra_options_pre_init(self, vim_config):
```
SirVer: This is now called _extra_vim_config in master.

doc/UltiSnips.txt
```diff
@@ -302,7 +303,33 @@ then you define your mapping as >
...
+3.2.2 Inner mappings                          *UltiSnips-inner-mappings*
+--------------------
+
+                                *UltiSnipsMapInnerKeys* *UltiSnipsUnmapInnerKeys*
+For maximum compatibility with other plug-ins, UltiSnips sets up some special
+mappings when a snippet starts being expanded, and tears them down once
+expansion is completed. In order to make it possible to override these inner
+mappings, it fires the following "User" autocommands:
+
+UltiSnipsMapInnerKeys
```
SirVer: How about UltiSnipsEnterFirstSnippet and *ExitLastSnippet?

UltiSnips now also sets up autocommands in these, so reducing this to mapping keys is no longer correct

doc/UltiSnips.txt
```diff
@@ -302,7 +303,33 @@ then you define your mapping as >
 and if the you can't expand or jump from the current location then the
 alternative function IMAP_Jumpfunc('', 0) is called.
 
-3.2.2 Path to Python module                   *UltiSnips-python-module-path*
+3.2.2 Inner mappings                          *UltiSnips-inner-mappings*
+--------------------
+
+                                *UltiSnipsMapInnerKeys* *UltiSnipsUnmapInnerKeys*
+For maximum compatibility with other plug-ins, UltiSnips sets up some special
+mappings when a snippet starts being expanded, and tears them down once
```
SirVer: `once expansion is completed` is inprecise. Maybe once the last snippet has been exited?

doc/UltiSnips.txt
```diff
@@ -302,7 +303,33 @@ then you define your mapping as >
...
+
+For example, to call a pair of custom functions in response to these events,
+you might do: >
+
+   autocmd! User UltiSnipsMapInnerKeys
+   autocmd User UltiSnipsMapInnerKeys call CustomInnerKeyMapper()
+   autocmd! User UltiSnipsUnmapInnerKeys
+   autocmd User UltiSnipsUnmapInnerKeys call CustomInnerKeyUnmapper()
+
+Note that because snippet expansion may be nested, it is possible that
```
SirVer: I think this is a unsymmetry that can be avoided with my suggested names up there.

test/test_Autocommands.py
```diff
@@ -0,0 +1,31 @@
+# encoding: utf-8
+from test.vim_test_case import VimTestCase as _VimTest
+from test.constant import *
+
+# Autocommands  {{{#
+
+class Autocommands(_VimTest):
```
SirVer: I think you want a more complex test too - a snippet with more than $0 so that you can expand a recursive snippet in it too.

SirVer:
Sorry for coming at you again with more comments. I worked on UltiSnips somewhat today, so my context is a little fresher. So more stuff jumped at me. What do you think?

doc/UltiSnips.txt
wincent: Done.

doc/UltiSnips.txt
wincent: I've updated the names, although see my comment below.

doc/UltiSnips.txt
wincent: Am I misreading the implementation? `_setup_inner_state` and `_teardown_inner_state` look asymmetrical to me, which implies the names should be `UltiSnipsEnterSnippet` (not `EnterFirst`) and `UltiSnipsExitLastSnippet`).

`_setup_inner_state` gets called repeatedly (and runs) when nesting snippets, but `_teardown_inner_state` has a guard (via `_current_snippet_is_done`) that prevents it from being called until the last snippet it exited.

(See the updated tests that I just pushed to https://github.com/wincent/ultisnips/commits/autocommands-squashed that show that we `_setup_inner_state` gets called twice for a snippet nested inside a snippet, but `_teardown_inner_state` only gets called once.)

It would be nice to have symmetrical names, but then they'd be not totally accurate because the methods aren't actually symmetrical.

test/test_Autocommands.py
wincent: I've updated this to be nested/recursive.

test/test_Autocommands.py
wincent: Thanks for the heads-up.

wincent:
So I rebased and squished all the commits into one. The new branch containing the single squashed commit is:

https://github.com/wincent/ultisnips/commits/autocommands-squashed

doc/UltiSnips.txt
SirVer: I fixed the method on master to be symmetric in 29d7080d64bb35be3f4aee9d7bd60d26679f6fdb. I dunno why they weren't before.

SirVer:
I made the methods symmetric on master. Could you update your code for this symmetry? I think than it is good to go.

Please add yourself to the list of contributors in doc/Ultisnips.txt at the very end if you have not yet.

wincent:
Ok, name updated to `UltiSnipsEnterFirstSnippet`, and updated the contributors list.

Commit is:
https://github.com/wincent/ultisnips/commit/9dd2a703653890a641fde20ae981d451bac2c3d7

SirVer:
Percet. Cherry picked on master as c4ee3ecf10ac942abe861aed5816b1628bda7b90. I also added a warning that it is unwise to change the buffer in any way in the commit following this.

Thanks for your work on this, wincent!

PR #522, opened 2015-06-27
VimBuffer: use vim.current.buffer.number instead of vim.eval

blueyed:
This should be more performant.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/522)


SirVer:
Thanks!

PR #537, opened 2015-07-31
Pipe-quoting helptag reference to UltiSnips-context-snippets

krisdages:
Was getting a "duplicate tag" error when running Helptags

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/537)


SirVer:
thanks :)

PR #538, opened 2015-08-01
To Fix E154: Duplicate tag "UltiSnips-context-snippets"

byaruhaf:
Error in how the documentation was formatted. Like the Error is saying, duplicate tags are detected. Vim uses asterisk surrounded text for its help tags. i kept getting this Error during plugin update with vundle. Fixed by removing the duplicate tag's

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/538)


SirVer:
Fixed in https://github.com/SirVer/ultisnips/pull/537 already. Thanks though.

PR #539, opened 2015-08-03
Autotrigger

seletskiy:
Guess what? It's me again.

This feature is a logical continuation of two preceding: context snippets and snippet actions.

Autotrigger combines them to achieve even faster programming style: there are
many places in a program code where only specific constructions can occurs.
E.g., in Go code on the top level you can see mostly (**p**)ackage / (**t**)ype / (**f**)unction / (**m**)ethod / (**v**)ar / (**c**)onst definition and nothing more.

So why use `p/t/f/m/v/c <TAB>` to expand snippet if one can use `p/t/f/m/v/c` alone which cause specific snippet to expand?

There are numerous applications for this feature, for example expanding operators in the middle of the code, e.g.
```
if (a =|)  ->  if (a == |)    automatically convert = to ==<SPACE>
```

I'm aware that it can be slow, I have some thoughts about optimization, but right now I'm using it and not experiencing any significant slowdown.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/539)

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -144,3 +144,8 @@ endf
 function! UltiSnips#LeavingInsertMode()
     exec g:_uspy "UltiSnips_Manager._leaving_insert_mode()"
 endfunction
+
+function! UltiSnips#TrackChange()
+    exec g:_uspy "UltiSnips_Manager._track_change()"
+endfunction
+" }}}
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1606,7 +1606,7 @@ endsnippet
 4.11 Autotrigger                                       *UltiSnips-autotrigger*
 ----------------
 
-Many language constructions can occur only at specific places, so it's
+Many language constructs can occur only at specific places, so it's
 possible to use snippets without manually triggering them.
 
 Snippet can be marked as autotriggered by specifying 'A' option in the snippet
@@ -1615,6 +1615,9 @@ definition.
 After snippet is defined as being autotriggered, snippet condition will be
 checked on every typed character and if condition matches, then snippet will
 be triggered.
+
+*Warning:* using of this feature can lead to significant vim slowdown. If you
+discovered that, report an issue to the github.com/SirVer/UltiSnips.
 
 Consider following snippets, that can be usefull in Go programming:
 ------------------- SNIP -------------------
@@ -1630,7 +1633,8 @@ endsnippet
 ------------------- SNAP -------------------
 
 When "p" character will occur in the beginning of the line, it will be
-automatically expanded into "package main". Same with "m" character.
+automatically expanded into "package main". Same with "m" character. There is
+no need to press trigger key after "m".
 
 ==============================================================================
 5. UltiSnips and Other Plugins                      *UltiSnips-other-plugins*
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1606,6 +1606,8 @@ endsnippet
 4.11 Autotrigger                                       *UltiSnips-autotrigger*
 ----------------
 
+Note: vim should be newer than 7.4.214 to support this feature.
+
 Many language constructs can occur only at specific places, so it's
 possible to use snippets without manually triggering them.
 
@@ -1634,7 +1636,7 @@ endsnippet
 
 When "p" character will occur in the beginning of the line, it will be
 automatically expanded into "package main". Same with "m" character. There is
-no need to press trigger key after "m".
+no need to press trigger key after "m""
 
 ==============================================================================
 5. UltiSnips and Other Plugins                      *UltiSnips-other-plugins*
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -42,6 +42,7 @@ UltiSnips                                      *snippet* *snippets* *UltiSnips*
       4.10.1 Pre-expand actions                 |UltiSnips-pre-expand-actions|
       4.10.2 Post-expand actions                |UltiSnips-post-expand-actions|
       4.10.3 Post-jump actions                  |UltiSnips-post-jump-actions|
+   4.11 Autotrigger                             |UltiSnips-autotrigger|
 5. UltiSnips and Other Plugins                  |UltiSnips-other-plugins|
    5.1 Existing Integrations                    |UltiSnips-integrations|
    5.2 Extending UltiSnips                      |UltiSnips-extending|
@@ -689,6 +690,9 @@ The options currently supported are: >
        controlled not only by previous characters in line, but by any given
        python expression. This option can be specified along with other
        options, like 'b'. See |UltiSnips-context-snippets| for more info.
+
+   A   Snippet will be triggered automatically, when condition matches.
+       See |UltiSnips-autotrigger| for more info.
 
 The end line is the 'endsnippet' keyword on a line by itself. >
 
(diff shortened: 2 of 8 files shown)


SirVer:
> Guess what? It's me again.

Yai!!! Welcome back :)

This feature was requested before, so I think that is a good use of time. The tests are sick though - and not only due to flakyness it seems.

---

Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks broke.

---

<sup>**[doc/UltiSnips.txt, line 1609 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwJV_kCXxk_gmSwyzD)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/doc/UltiSnips.txt#L1609)):</sup>
language constructs

---

<sup>**[doc/UltiSnips.txt, line 1617 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwJb5p2TwiGVrg2ACs)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/doc/UltiSnips.txt#L1617)):</sup>
Warn about the potential slowdown in Vim for running a command on each keystroke.

---

<sup>**[doc/UltiSnips.txt, line 1632 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwJjLuyh2W00ld0e-d)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/doc/UltiSnips.txt#L1632)):</sup>
mention that there is no need to press the trigger key after the m.

---

<sup>**[pythonx/UltiSnips/snippet/source/_base.py, line 31 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwJxwXTNHJZCqF4L3y)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/pythonx/UltiSnips/snippet/source/_base.py#L31)):</sup>
Why not do the filtering further up instead of handing autotrigger_only all the way down?

---

<sup>**[pythonx/UltiSnips/snippet/source/_base.py, line 43 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwK8SPVTYCZN0pNqWu)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/pythonx/UltiSnips/snippet/source/_base.py#L43)):</sup>
> 80 chars

---

<sup>**[pythonx/UltiSnips/snippet/source/_snippet_dictionary.py, line 19 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwKAotdzOVBu7UHwNb)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/pythonx/UltiSnips/snippet/source/_snippet_dictionary.py#L19)):</sup>
document `autotrigger_only` in the docstring. Also mention why it is handed down instead of filtering after the function has been called.

---

<sup>**[pythonx/UltiSnips/snippet_manager.py, line 778 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwKVMGEqjo2lGoiwVM)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/pythonx/UltiSnips/snippet_manager.py#L778)):</sup>
I guess this function is a opitmization? I mean, CursorMoved used to do something similar, so you could reuse that code. 

---

<sup>**[test/test_Autotrigger.py, line 7 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwKgpGXJatpQ5HOGFK)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/test/test_Autotrigger.py#L7)):</sup>
Can you add a test with a len(trigger) > 1 

---

<sup>**[test/test_UltiSnipFunc.py, line 158 \[r1\]](https://reviewable.io:443/reviews/sirver/ultisnips/539#-JvwKnInR4PzrHXLsj0a)** ([raw file](https://github.com/sirver/ultisnips/blob/6c4fb6f3f21f03acf0a93acdb464be87e799d26c/test/test_UltiSnipFunc.py#L158)):</sup>
`and not autotrigger_only` ? Comparing a bool with a bool to get a bool always feels weird to me.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/539)

seletskiy:
@SirVer:

>  I guess this function is a opitmization? I mean, CursorMoved used to do something similar, so you could reuse that code. 

`UltiSnips#CursorMoved` is triggered only on `CursorMoved/CursorMovedI`. `track_change` function is triggered on `InsertCharPre/TextChangedI`, so it's completely different.

>  and not autotrigger_only ? Comparing a bool with a bool to get a bool always feels weird to me.

`and not autotrigger_only` sounds even more weird for me. Like double negation, so confusing.

Regarding failed tests: there is the problem with vanilla unpatched 7.4 vim, I don't know the roots of this problem, but I've got the exception while expanding snippet in `track_change`, which is triggered by autocommand. Something about `Not allowed here` on buffer modification from python code.
It doesn't reproduce somewhere after patch 214 (not exactly, it's lowest version of pre-built package which I have that works with autotrigger). I do not have time nor motivation to dig down the problem with outdated vim, so I've just fixed tests by adding ugly patch version comparison.

Everything else should be fixed right now.

SirVer:
Vim 7.4 is not outdated - it is the latest officially released version of the editor. And it is the one that UltiSnips supports officially. So you at least need to provide a user warning when the feature is used on Vims prior to that patch level - or you find a way to make it work.

Merging it in a state where it bugs out on old Vims leads to bug reports and confused users.

seletskiy:
It's outdated because there are more than 800 patches available which are officially accepted by maintainer.

There will be no error or exception and I clearly stated minimal required version in the first line of documentation on this feature.

SirVer:
> It's outdated because there are more than 800 patches available which are officially accepted by maintainer.

that counts for very little - as long as there is no 7.4.1 or 7.5 available, 7.4 vanilla is the latest official version and the one we have to support. For example, Ubuntu 14.4 LTS currently ships with vim 7.4 patches 1-52. 

> There will be no error or exception and I clearly stated minimal required version in the first line of documentation on this feature.

Could you clarify - I understood you earlier that tests threw errors on  the tests throw errors on vanilla 7.4 - I see that you work around that by checking for the version in the test. What happens if a user with an older 7.4 tries to use your feature? My understanding is that an error is thrown and this would likely result in spurious bug reports on UltiSnips. Is my reasoning flawed?

seletskiy:
> Could you clarify - I understood you earlier that tests threw errors on the tests throw errors on vanilla 7.4 - I see that you work around that by checking for the version in the test. What happens if a user with an older 7.4 tries to use your feature? My understanding is that an error is thrown and this would likely result in spurious bug reports on UltiSnips. Is my reasoning flawed?

Yes. Tests was broken not because of exception or something. They are broken because expected output is not met. So snippets simply will not be autotriggered. I can add testcase for the vim < 7.4.214 that will show, that there is no exception.

seletskiy:
@SirVer: BTW, something wrong with neovim tests in travis. Everything works on my local machine.

seletskiy:
> I can add testcase for the vim < 7.4.214 that will show, that there is no exception.
Added.

SirVer:
code lgtm. I am not sure what is breaking on travis. It runs on my local machine just fine. I'll merge and set the neovim tests to allow fail till we can figure that out. 

---

Reviewed 2 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sirver/ultisnips/539)

seletskiy:
@SirVer: Great! :+1:

Issue #542, opened 2015-08-06
UltiSnips leave trigger text when expanding snippets when filetype is set in modeline


seletskiy:
I've found bug which was introduced in commit c4ee3ec. Steps to reproduce:

* create file `/tmp/testfile`;
* write following to `/tmp/testfile`:
  ```
  # vim: ft=sh

  ```

* create snippet for `sh` filetype, e.g. (doesn't matter, actually):
  ```
  snippet a "alias" b
  alias
  endsnippet
  ```

* try to expand `a` in the `/tmp/testfile`:
  ```
  # vim: ft=sh
  a<TAB>
  ```

* see:
  ```
  # vim: ft=sh
  aalias
  ```

Commenting out following line (in the `pythonx/UltiSnips/snippet_manager.py`) will resolve issue:
```py
_vim.command('doautocmd User UltiSnipsEnterFirstSnippet')
```

I tried to define noop autocommand in my `.vimrc` for the `UltiSnipsEnterFristSnippet`, but it didn't help.

Also, this issue somehow linked with YouCompleteMe. Disabling YouCompleteMe resolves issue as well.

@wincent, you are probably interested in this issue.

wincent:
I can repro this (thanks for all the detail). I think we want `<nomodeline>`; that resolves the issue:

```
_vim.command('doautocmd <nomodeline> User UltiSnipsEnterFirstSnippet')
```

From `:h :doautocmd`:

```
							*<nomodeline>*
			After applying the autocommands the modelines are
			processed, so that their settings overrule the
			settings from autocommands, like what happens when
			editing a file. This is skipped when the <nomodeline>
			argument is present. You probably want to use
			<nomodeline> for events that are not used when loading
			a buffer, such as |User|.
```

I'll whip up a patch.

PR #544, opened 2015-08-07
Use <nomodeline> when firing user autocmds

wincent:
As recommended by the Vim help (`:h <nomodeline>`):

    After applying the autocommands the modelines are
    processed, so that their settings overrule the
    settings from autocommands, like what happens when
    editing a file. This is skipped when the <nomodeline>
    argument is present. You probably want to use
    <nomodeline> for events that are not used when loading
    a buffer, such as |User|.

Fixes: https://github.com/SirVer/ultisnips/issues/542

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/544)

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -374,7 +374,7 @@ class SnippetManager(object):
             'autocmd CmdwinLeave * call UltiSnips#LeavingBuffer()')
         _vim.command('augroup END')
 
-        _vim.command('silent doautocmd User UltiSnipsEnterFirstSnippet')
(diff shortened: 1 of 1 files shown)


SirVer:
Awesome! Could you add yourself to the authors list in the documentation?

SirVer:
pff - sorry, disregard that. too early in the morning :/

wincent:
Thanks for merging so quickly @SirVer.

Looking at `:h version7.txt`, I see that `<nomodeline> ` was only added to Vim in 7.3.438, so we should probably make this apply conditionally. I'll push a follow-up in a minute.

PR #545, opened 2015-08-07
Only use <nomodeline> if we're sure it is available

wincent:
According to `:h version7.txt`, `<nomodeline>` was added in Vim 7.3.438.
Trying to use it prior to that will likely result in an:

    E216: No such group or event: <nomodeline> User UltiSnipsEnterFirstSnippet

Add a conditional to ensure we only use it when we know it's available.

Follow-up to: https://github.com/SirVer/ultisnips/pull/544

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/545)


SirVer:
Ultisnips requires vim 7.4 as to the docu. Vim 7.3 is not supported. I think this can be dropped.

wincent:
Cool! Good to know.

PR #548, opened 2015-08-10
Remove autoload/UltiSnips/bootstrap.vim again

blueyed:
This seems to have been re-added accidentally in 3c4ac11, after it was
removed in becab78.

It is referred to in `test.py` btw.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/548)


SirVer:
Thanks Daniel, always on point.

seletskiy:
Oops, sorry.

PR #549, opened 2015-08-10
Define UltiSnips#FileTypeChanged always

blueyed:
This is called via ftdetect/UltiSnips.vim, and needs to be defined
also if UltiSnips fails to initialize, e.g. because of missing Python
support.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/549)

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -1,5 +1,9 @@
 if exists("b:did_autoload_ultisnips") || !exists("g:_uspy")
-   finish
+    " Define no-op function, called via ftdetect/UltiSnips.vim.
+    function! UltiSnips#FileTypeChanged()
+    endfunction
+
+    finish
 endif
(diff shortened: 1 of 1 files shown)


SirVer:
I feel that this regressed recently -  I think we had such a dummy function around and I might have removed it in a refactoring. Do you have an idea of how to test that this does not regress again?

blueyed:
> Do you have an idea of how to test that this does not regress again?

Passing in / using an invalid Python version should trigger it, via `g:_uspy` not being set.
And then the test would need to trigger the FileType autoevent, and should not throw an error because of the missing function.
The test might also check for the function to exist, but I think a general test to check if inserting and leaving insert mode works is better.

SirVer:
Good idea with the wrong python version setting.

I have no time for working on this right now, but I added a TODO while merging your work so that it can be done later. 

Thanks for the patch!

PR #615, opened 2015-11-28
Updated changelog, readme and documentation for 3.1 and change of ownership.

SirVer:
I made the documentation shorter and removed some fluff. I removed all uses of 'me' and mentioned the change of maintenance in appropriate places. Once this is merged I will do the 3.1 release and be on my way.

@seletskiy Could you have a look if you agree with everything in this PR?

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/615)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,23 +1,32 @@
-changes up to 3001566:
+version 3.1 (TODO: date):
+	- This is the last release done by @SirVer. The new maintainer of UltiSnips
+	  is @seletskiy. The repository remains https://github.com/SirVer/ultisnips,
+	  so this should not affect any users. This is also the last release to be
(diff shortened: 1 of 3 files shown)


seletskiy:
@SirVer: Is it OK that travis tests failed for python 2.7?

seletskiy:
Otherwise, look good.

SirVer:
The travis test failed due to flakyness. I usually just restart it then. I think you should be able to restart it yourself according to https://docs.travis-ci.com/user/travis-pro/#Who-has-access-to-the-builds%3F. That is all the access I can grant on travis - there does not seem to be a configuration option for more rights. 

I am going to merge this now. I will check out of UltiSnips soonish then.

PR #625, opened 2015-12-21
add snippets_in_current_buffer function

chemzqm:
Add a function list all available snippets of current file.

The result is a list which also have location information, so that the snippet could be previewed in unite.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/625)

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -93,7 +93,7 @@ function! UltiSnips#SnippetsInCurrentScope(...)
     let g:current_ulti_dict = {}
     let all = get(a:, 1, 0)
     if all
-      let g:current_ulti_dict_all = {}
+      let g:current_ulti_dict_info = {}
     endif
     exec g:_uspy "UltiSnips_Manager.snippets_in_current_scope(" . all . ")"
     return g:current_ulti_dict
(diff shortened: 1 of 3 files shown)


seletskiy:
Thanks for the contribution!

I guess we do not need another method with exactly the same code as in `snippets_in_current_scope`.
So I propose to slightly modify `snippets_in_current_scope` code so it will accept `before` argument with default value of `_vim.buf.line_till_cursor` and will return `dict` with additional values that you expect.

chemzqm:
@seletskiy sure, that would be nice.
A problem is currently the returned dict only have key value for trigger and description,
don't know how to implement it with additional values without change the API.

seletskiy:
@chemzqm: Ah, I see. I guess you can safely add another dict assignment into separate variable right after `g:current_ulti_dict` assignment.

seletskiy:
@chemzqm: ping?

chemzqm:
Working on it, I'll send a PR soon.

chemzqm:
I've made a optional `all` param for UltiSnips#SnippetsInCurrentScope function which is default `0`
so there is would be no conflict with existing code,  user could get all snippets of current buffer with location
like this:

``` VimL
call UltiSnips#SnippetsInCurrentScope(1)
echo g:current_ulti_dict_all
```

seletskiy:
@chemzqm: Sorry for the long delay, I've had intense schedule recently.

Great work, thanks for the contribution! Expecting to see new contributions from you :+1:

PR #628, opened 2015-12-30
Fix highlighting for nested visual tabstob

kmarius:
Consider the snippet

snippet test
${1:${VISUAL}}
endsnippet

A missing 'extend' keyword caused the inner closing bracket to end both
the inner 'snipVisual' and the outer 'snipTabStop' regions. Anything
after the inner bracket was considered 'snipSnippetBody' and thus
highlighted incorrectly.

[Review on Reviewable](https://reviewable.io/reviews/sirver/ultisnips/628)


seletskiy:
@kmarius: Great! Thanks! :cake:

PR #647, opened 2016-01-22
Allow Deoplete entries to show up with existing matches

alexlafroscia:
I encountered an issue with Deoplete where Ultisnips suggestions wouldn't appear if there was a match in the current buffer with the same name as a Ultisnips snippet.  This change allows the snippet and buffer match to appear side-by-side, which is less confusing.

More information can be found on the issue on the Deoplete repo here:

https://github.com/Shougo/deoplete.nvim/issues/138#issuecomment-174050309


megalithic:
:+1:

seletskiy:
@alexlafroscia: thx! :cake:

PR #648, opened 2016-01-23
Expand environment variables in UltiSnipsSnippetsDir

alexlafroscia:
I noticed that environment variables don't work in the current version of UltiSnips within the value for `g:UltiSnipsSnippetsDir`.  For example, I tried setting it to

```vim
let g:UltiSnipsSnippetsDir = "$DOTFILES/nvim/UltiSnips"
```

But it would append the directory after the path to the current file or something weird like that.  This change expands the path at the right time so that it can find the snippet directory correctly.

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -795,8 +795,8 @@ class SnippetManager(object):
             if bang:
                 potentials.update(find_all_snippet_files(ft))
 
-        potentials = set(os.path.realpath(os.path.expanduser(p))
-                         for p in potentials)
+        potentials = set(os.path.realpath(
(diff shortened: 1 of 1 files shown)


seletskiy:
@alexlafroscia: Why not just

```
let g:UltiSnipsSnippetsDir = $DOTFILES . "/nvim/UltiSnips"?
```

alexlafroscia:
Huh, I suppose that works too 😝 for some reason that didn't cross my mind. I'm fine with closing this PR with merging in favor of that being the correct way to achieve that instead.

seletskiy:
@alexlafroscia: So, I'm closing PR in favor of using simpler

alexlafroscia:
:+1:

PR #684, opened 2016-04-13
allow specify context using special header

kovetskiy:


seletskiy:
That's great, thank you!

PR #686, opened 2016-04-14
UltiSnipsEdit should try to find non-empty snippet files

kovetskiy:


seletskiy:
@kovetskiy: Thanks for the great contribution!

PR #720, opened 2016-07-07
Include examples about how to create postfix snippets

thalesmello:
Postfix snippets are an incredibly useful feature in IntelliJ IDE's.

Only recently I came to realize this [could be implemented with UltiSnippets](https://github.com/sniphpets/sniphpets-postfix-codes/blob/master/UltiSnips/php.snippets).

I think it would be a nice addition to the UltiSnippets documentation, in the hopes it could help others spend less time than I did trying to figure out how to put it together.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1010,6 +1010,26 @@ expressions in this manner has some drawbacks:
    would end up with the above SNAP instead, not what you want.
 2. The snippet is harder to read.
 
+The biggest advantage, however, is that you can create snippets that take into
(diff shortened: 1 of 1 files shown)


seletskiy:
@thalesmello: It's good, but think about the evolution of that pattern: usually, you want to wrap a current line into specific expression depending on the context. Like, usually you want to insert `return` if current line is at the end of the current code block. So, you may ease your typing by making smart snippet, which will match not just `blah;ret<TAB>`, but just `blah;<TAB>` only if `blah;` is at the end of code block. See more about how you can achieve this in the `:h UltiSnips-context-snippets`.

PR #735, opened 2016-08-16
fix visual placeholder multi call

SolaWing:
in `exec "xnoremap <silent> " . g:UltiSnipsExpandTrigger. " :call UltiSnips#SaveLastVisualSelection()<cr>gás"`

mapping to `:call` will actually execute `:'<,'>call`, which will call the function multiple times.  This will cause delay when visual contains many lines, and may have other unexpected behavior. so add the range argument


seletskiy:
@SolaWing: Thanks!

PR #745, opened 2016-09-02
Improve/simplify filetype handling for buffers

blueyed:
There is no need to have a FileType autocmd anymore, which for some odd
reason even had to be in "ftdetect"?!

Now it will just use the `&filetype` property from buffers, instead of
updating it on FileType events.

This allows for easier lazy-loading of UltiSnips; without this patch it
would not have updated the list of filetypes for the buffer after just
being triggered (e.g. through NeoBundle's lazy-loading).
I am using the following currently:

    NeoBundleLazy 'SirVer/ultisnips', {
      \ 'on_funcs': ['UltiSnips#ExpandSnippetOrJump']}
    inoremap <silent> <c-j> <C-R>=UltiSnips#ExpandSnippetOrJump()<cr>

By manually defining the mapping for the trigger, I can invoke it and
NeoBundle will trigger the loading of the plugin.

NOTE: I've been using this since a while already, but it had two conflicts when rebasing - so take it with care.


seletskiy:
@blueyed: So it should not basically affect previous behaviour, isn't it?

seletskiy:
@blueyed: I'm getting following error when trying to hit tab on file without filetype using your path:

```
An error occured. This is either a bug in UltiSnips or a bug in a
snippet definition. If you think this is a bug, please report it to
https://github.com/SirVer/ultisnips/issues/new.

Following is the full stack trace:
Traceback (most recent call last):
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/err_to_scratch_buffer.py", line 16, in wrapper
    return func(self, *args, **kwds)
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 124, in expand
    if not self._try_expand():
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 702, in _try_expand
    snippets = self._snips(before, False, autotrigger_only)
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 575, in _snips
    source.ensure(filetypes, cached=autotrigger_only)
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet/source/file/_base.py", line 47, in ensure
    self._load_snippets_for(ft)
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet/source/file/_base.py", line 80, in _load_snippets_for
    self._parse_snippets(ft, fn)
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet/source/file/_base.py", line 93, in _parse_snippets
    file_data = compatibility.open_ascii_file(filename, 'r').read()
  File "/home/s.seletskiy/.vim/bundle/ultisnips/pythonx/UltiSnips/compatibility.py", line 76, in open_ascii_file
    return open(filename, mode)
IOError: [Errno 21] Is a directory: '/var'
```

blueyed:
@seletskiy 
Just don't do that.. ;)

I've pushed a fixup.  (NOTE: this should be squashed and not get merged as-is)

blueyed:
> So it should not basically affect previous behaviour, isn't it?

Yes.

seletskiy:
@blueyed: Looks lgtm for me. Can you, please, squash commits then?

seletskiy:
@blueyed: Never mind, GitHub did it for us.

Thanks for the contribution and be in touch!

myii:
@blueyed I haven't got around to upgrading my Vim plugins for a while.  When I came around to `ultisnips`, it was no longer working.  YouCompleteMe wasn't bringing up the snippets list.  Worked through the commits until I determined that it was this PR that made the difference.

I've been able to revert 8d77e89 on top of the [current `master` branch](https://github.com/SirVer/ultisnips/tree/6fdc3647f72e0a1f321ea6bd092ecd01f7c187ba0) and it is working as usual again.

Alas, I don't have the time to investigate what part of this commit is causing the issue.

blueyed:
@myii 
The comments with that commit indicate that it might be missing a YCM update?  But it's really old already..

myii:
@blueyed Appreciate your quick response.  I didn't notice the comment about YCM but I'll look into that when I get a chance.

myii:
@blueyed So I bit the bullet and upgraded YCM, including all of the hassle.  But it works now, without the revert.  Thanks for your advice and apologies for the noise.

blueyed:
@myii 
Phew, I am glad it worked out.
Enjoy your updated software! :)

PR #844, opened 2017-04-19
Mark Deoplete source as volatile

tombeynon:
As per https://github.com/Shougo/deoplete.nvim/pull/468#issuecomment-295168539 - UltiSnips suggestions depend on user input so the Deoplete source should be marked as `is_volatile`. Without this option Deoplete can cache an empty result from UltiSnips, causing no further snippets to be shown when the user input changes.

### DIFF
--- a/rplugin/python3/deoplete/sources/ultisnips.py
+++ b/rplugin/python3/deoplete/sources/ultisnips.py
@@ -7,6 +7,7 @@ class Source(Base):
         self.name = 'ultisnips'
         self.mark = '[US]'
         self.rank = 8
+        self.is_volatile = True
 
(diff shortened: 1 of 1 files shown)


tombeynon:
To explain this a bit further - Deoplete expects to carry out the filtering of snippets by default, but the UltiSnips source uses `UltiSnips#SnippetsInCurrentScope()` which filters the returned snippets based on user input. Without `is_volatile`, Deoplete will cache the first set of results UltiSnips returns and keep trying to filter that, rather than getting a new set of results when the input changes.

SirVer:
A straightforward fix. Thanks @tombeynon

Issue #822, opened 2017-02-12
UltiSnips refuses to move to next tab stop after using YouCompleteMe semantic completion


ghost:
I am using UltiSnips together with YouCompleteMe, for writing C/C++ code. Both of the plugins work great, but whenever I am in the middle of a snippet (editing one of the tab stops), and I use YCM to, say, complete a field inside a struct, UltiSnips refuses to move to the next tab stop after that (maybe it considers the snippet complete??)

I am using `<tab>` for YouCompleteMe, and F12 as my key for UltiSnips.

To illustrate the problem, say I have a snippet like this:

```
snippet sc "strcmp"
strcmp($1, $2)
endsnippet
```

If I wanted to use it to write this:

```
strcmp(hello, world)
```

everything works fine. I type `sc`, press F12, snippet is expanded, cursor moves to `$1`, I write `hello`, press F12, cursor moves to `$2`, and I write `world`.

However, if I wanted to write something like:

```
strcmp(mystruct->label, other_label)
```

what happens is the following: I type `sc`, press F12, snippet is expanded, cursor moves to `$1`, I write `mystruct->`, the YCM completion menu shows up, I press `<tab>` to complete `label`, and then I want to move to the next tab stop, so I press F12 again, but nothing happens. The cursor stays where it is.

The problem only seems to occur with YCM completion menus for struct fields and such, and not with regular keyword completion; i.e I can type `co` and press `<tab>` to autocomplete `const`, and then UltiSnips still works fine, but when I am using semantic completion from YCM, UltiSnips breaks. Also,
when the YCM completion menu appears after I type `->`, I can still press F12 and move to the next tab stop successfully, as long as I haven't yet pressed `<tab>` to complete the struct field. Hence, the problem seems to be triggered by the completion operation when I press `<tab>`, rather than the displaying of the YCM completion menu.

My UltiSnips configuration options in my `~/.vimrc` are:
```
let g:UltiSnipsExpandTrigger="<F12>"
let g:UltiSnipsJumpForwardTrigger="<F12>"
let g:UltiSnipsJumpBackwardTrigger="<s-F12>"
```

I have not set any configuration options for YCM.

I have no idea what is causing this or how to fix it. Any ideas will be appreciated.

P.S: I didn't know if it was more appropriate to report this issue on the UltiSnips issues page or on the YCM issues page, so I am just posting it here.

mseabold:
[...]

PR #845, opened 2017-04-20
Add option to disable preview window in a snippet

### LINKED ISSUE #681: ultisnips forgets tabstops when preview window opens
I'm using neovim with deoplete, deoplete-jedi, and ultisnips (all latest versions from git) to edit a python file. Jedi pops up the preview window to show a docstring for the function I'm editing, but then any snippet I was writing suddenly loses the ability to jump through tabstops after the preview window pops up.

mseabold:
This is in order to workaround YCM (#822) when it has the preview enabled. If
YCM triggers the preview while in a snippet, the current snippet context
is torn down.

In the future it may be useful to gracefully maintain the state in this
case, but for now I believe this is a reasonable workaround.

I made this behavior disabled by default.

Fixes #681.


ThePlasmaRailgun:
This seems pretty useful... I've been looking for a fix for this.

puremourning:
I have had a customisation like this for a very long time, due to the bad interraction with the preview window. It works well. so 👍 to this.

SirVer:
@mseabold Sorry for the long delay in reviewing. This needs documentation update and a test. Are you up for adding it?

mseabold:
I'll have to familiarize myself with the test infrastructure to figure out how to effectively test this, but I should be able to do it.

SirVer:
I hope to add documentation for this soonish. In the meantime I suggest starting with the Travis yml file.

SirVer:
@mseabold There is now a one pager about how testing in UltiSnips works: https://github.com/SirVer/ultisnips/blob/master/CONTRIBUTING.md. Please let me know if it is understandable and suggest changes otherwise.

mseabold:
I think that document is pretty clear. I just need to come up with a case that would normally trigger the preview. I'll get it done when I get chance soon

DaveDay:
I too would love a fix for this issue, so much so that today I went ahead and applied the patch locally.  Unfortunately, I see no change in behavior at all.
I have included ``let g:UltiSnipsDisablePreview = 1
just after my "Plug 'Sirvir/Ultisnips'" invocation.
Is that the correct way to set that variable?

Thanks,
Dave Day

mseabold:
Dave,

I'm not totally sure without more details to recreate it. Perhaps setting that inside plug#begin() is not kosher?

DaveDay:
mseabold, 
Thanks for your support.  I had 
completeopt=menu,menuone,preview,noinsert
set near the top of my vimrc.  When I commented that out
so that nowhere in my configs were completeopt mentioned,
it started working correctly.
I don't really know why, since I understood the patch to be saving and then
restoring the completeopt.  

In any case, it is working now, though I haven't had enough time to verify that all
other behaviors are unchanged.

Thanks so much.
Dave Day

mseabold:
It should be saving and restoring but I'll try to put together some test cases to make sure that's working correctly

DaveDay:
mseabold,
Thanks for your efforts...
After rereading my last post, I see I was  misleading when I said "it started working correctly", it did not.  What I should have said is that now the preview window NEVER opens just as you would expect if those two patches were not in the code, though they are.  Sorry for the confusion.  

So I have to guess that those patches are not being run somehow.  

I also tried disabling their "if" statements so that they would be run no matter what but no behavior changes were apparent.
I also just substituted _vim.command('set completeopt-=preview') 
for the first patch, and
_vim.command('set completeopt+=preview')
for the second, just to see what happened, and observed no behavior changes at all with anything.
It is almost like those patches are not being executed at all.

Cheers,
Dave

mseabold:
Dave,

I'm not able to produce any issues in my current setup. I changed my completeopt to match yours and I saw expeted behavior: Preview works outside of a snippet, preview is disabled during a snipper, preview works once snippet is completed

I'm not sure what plugins you're using or how you configured your vimrc

DaveDay:
Matt, 
It works so long as I do not set completeopt in my vimrc.  No idea why.

SirVer:
@mseabold Any update on this?

DavidUser:
Work's fine to me to solve the conflicts with YCM preview. But should have some way of save the context, to instead just disable preview, let preview works and back normally to the context when the next "tab" is pressed.

SirVer:
I found #1055 (which was merged today) to be a simpler fix for this problem. Closing this, since they should have addressed the same problems. If I am mistaken about this, please reopen and let me know.

PR #860, opened 2017-06-13
Fix emphasis in doc file that created a tag

jpfender:
Emphasizing 'not' as `*not*` causes `:helptags` to generate a tag called
'not'. Because people like writing `*not*` in their documentations, this
can cause duplicate tag conflicts with other plugins (e.g. Command-T's
doc also generates a 'not' tag).


seletskiy:
@jpfender: Thanks!

PR #864, opened 2017-06-23
Fix selecting one more character when using ${VISUAL}.

J3soon:
**This fix is for Windows only**
- Remove the last character in visual selection to fix the #863 issue.
- Fix typo in comment.

Issue #863, opened 2017-06-23
Selecting one more character when using ${VISUAL}


J3soon:
When selecting words in Vim's Visual-Mode, it will put one more character (which isn't selected) in the `${VISUAL}` placeholder.

**Demo:**

![vim-ultisnips](https://user-images.githubusercontent.com/20457146/27464386-d2c7e49a-57ff-11e7-8c03-63ecec5f5d39.gif)

I'm using Vim 8.0 on 64-bit Windows 10

The snippet I use is: 

    snippet bold "Bold in markdown"
    **${1:${VISUAL}}**
    endsnippet

J3soon:
This problem doesn't exist on 64-bit Ubuntu 16.04...

laxtiz:
when i use the commit before ur PR. all ok
after update ultisnips with ur PR, i will miss ends char when expand a snippet with a VISUAL placeholder.
i don't know why

![snip](https://user-images.githubusercontent.com/3883767/29489993-1d3b8c00-8561-11e7-8a81-82b267e9cd9c.gif)

gvim x64 8.0.678 on windows 10, with python 3.6.2

J3soon:
@laxtiz, the gif you posted seems to work fine.
The end char is the one with the black cursor on it.
Can you explain the problem more detailed?

J3soon:
The correct behavior is shown in [UltiSnips Screencast - Ep 3 - What's new in Version 2.0?](https://youtu.be/hKrrOevbrxY?t=235)

laxtiz:
I took two pictures from the gif. 
I guess I probably know the reason

this pic is your:
![pr1 141](https://user-images.githubusercontent.com/3883767/29500647-4c0977fa-8655-11e7-9d6d-c43af9aa2aad.png)

and my:
![my 6](https://user-images.githubusercontent.com/3883767/29500658-5ec37684-8655-11e7-8875-4f4649f8ca5e.png)

why your curosr style is **LINE** when gvim in visual mode.
see `:h guicursor` plz.

J3soon:
@laxtiz Oh, so that's the problem...
I'd changed my cursor style to block and tested it, I found out that the char with black cursor on won't be included when using `${VISUAL}`.
However, if you look closely when selecting a text, you'll see the black cursor is blinking, indicating that the char is not selected. The `ve` key combination can show that too.
I think the char with black cursor on isn't meant to be selected, correct me if I'm wrong.

laxtiz:
[...]

PR #868, opened 2017-06-30
Fix selection when using ${VISUAL} in line selection mode on Windows.

J3soon:
The #864 fix (issue #863 ) causes selection problem when using multi-line selection.
- Fix the last character issue when using line selection mode (`V`) on Windows.


seletskiy:
@J3soon: Thanks!

laxtiz:
so now i will miss ends char when i expand snippet on windows.
please rollback PR #864 and #868 .

PR #872, opened 2017-07-22
Avoid mapping Visual-Line Selection mode

### LINKED ISSUE #838: Conflicts with own `<tab>` maps in visual and/or select mode
I have my own hooks for `<tab>` in *visual* and *select* modes:
```viml
xnoremap <tab> <esc>
snoremap <tab> <esc>
```
**UltiSnipts** overwrites my `<tab>` map for *select* mode, so I disabled maps that came out-of-the-box:
```viml
let g:UltiSnipsExpandTrigger = "<Nop>"
let g:UltiSnipsListSnippets = "<Nop>"
```
But **UltiSnips** also trying to do `unmap` for my `<tab>` map even if it is set to different key, so I found a way to disable it too:
```viml
let g:UltiSnipsRemoveSelectModeMappings = 0
```
And I set this map that I found in **UltiSnips** sources:
```viml
inoremap <silent> <Tab> <C-R>=UltiSnips#ExpandSnippet()<CR>
```
But now if I press `<tab>` in insert mode without any snippet it doesn't insert indentation that normally happens as expected when I keep default settings of **UltiSnips**. My question is: how it's done? How I can copy that behavior with my own mapping avoiding conflicts with my `<tab>` maps?

blackheaven:
I think it more or less fix #838, I don't think Visual-Line Selection mode isn't useful and it's the mode that I'm unable to map.

### DIFF
--- a/autoload/UltiSnips/map_keys.vim
+++ b/autoload/UltiSnips/map_keys.vim
@@ -61,7 +61,6 @@ function! UltiSnips#map_keys#MapKeys()
         exec "inoremap <silent> " . g:UltiSnipsExpandTrigger . " <C-R>=UltiSnips#ExpandSnippet()<cr>"
         exec "snoremap <silent> " . g:UltiSnipsExpandTrigger . " <Esc>:call UltiSnips#ExpandSnippet()<cr>"
     endif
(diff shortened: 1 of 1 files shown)


blackheaven:
Well... no, sorry

PR #883, opened 2017-08-21
fix bug when vim option 'selection' is 'exclusive'

laxtiz:
here is a vim option ``selection``.
if the value is `exclusive`, the mark `'>` is one more then real visual selection.
so var `ebyte` or `ec` must subtract 1.

and that is not a windows bug.
because gvim on windows sourced `$VIMRUNTIME\mswin.vim` by default.
the line 17 of `mswin.vim` execute `:behave mswin` will change option `selection` to `exclusive`.

more deailt in issue #863

### DIFF
--- a/pythonx/UltiSnips/vim_state.py
+++ b/pythonx/UltiSnips/vim_state.py
@@ -126,6 +126,11 @@ class VisualContentPreserver(object):
         ec = byte2col(el, ebyte - 1)
         self._mode = _vim.eval('visualmode()')
 
+        # fix bug when vim option 'selection' is 'exclusive'.
+        if _vim.eval('&selection') == 'exclusive':
+            if sl != el and sbyte != ebyte:
+                ec -= 1
+
         _vim_line_with_eol = lambda ln: _vim.buf[ln] + '\n'
 
         if sl == el:
--- a/pythonx/UltiSnips/text_objects/_visual.py
+++ b/pythonx/UltiSnips/text_objects/_visual.py
@@ -15,7 +15,6 @@ from UltiSnips import _vim
 from UltiSnips.indent_util import IndentUtil
 from UltiSnips.text_objects._transformation import TextObjectTransformation
 from UltiSnips.text_objects._base import NoneditableTextObject
-import platform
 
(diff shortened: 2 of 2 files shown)


pythonx/UltiSnips/vim_state.py
```diff
@@ -126,6 +126,11 @@ def conserve(self):
         ec = byte2col(el, ebyte - 1)
         self._mode = _vim.eval('visualmode()')
 
+        # fix bug when vim option 'selection' is 'exclusive'.
+        if _vim.eval('&selection') == 'exclusive':
+            if not (sl == el and sbyte == ebyte):
```
laxtiz: don't move cursor when you press `v` into visual mode .
vim will select the char under cursor, but position of mark `'<` is equal `'>` . 
the effect is the same as `inclusive` mode. do not need to be reduced by 1.

asknet:
@laxtiz May I know if this pull request has any issues?

J3soon:
@asknet The fix worked well and should be merged. However, I'm not sure why the test checks have failed.

SirVer:
@j3soon @laxtiz Is this fixed through https://github.com/SirVer/ultisnips/pull/926 as well? Can this one be closed?

laxtiz:
@SirVer if prees v and select only one char, the mark `'>` is right.
The PR #926 can't fixed this.

SirVer:
Can you add a test for this? I do not understand the problem.

j3soon:
@SirVer I think this is fixed through #926.

laxtiz:
okay，maybe later






发自：ZUK Z2
在 Holger Rapp <notifications@github.com>，2018年4月1日 下午4:16写道：

Can you add a test for this? I do not understand the problem.

SirVer:
@laxtiz @j3soon I had a more proper look at this patch and added two tests to verify behaviour.

I think this patch is more correct than #926. This one actually makes UltiSnips use the same semantics for visual selection as Vim, while before we were preserving the wrong Visual content and then fixing its use later on by dropping the last character.

@laxtiz Thanks for the fix!

@UltiBot merge

SirVer:
@UltiBot merge

PR #917, opened 2017-12-07
Add support for editing snippets in a new tab

anhari:
Adds a `tabdo` option to `:UltiSnipsEditSplit` for opening snippet files with `:UltiSnipsEdit` in a new tab.

I originally had `tab` instead of `tabdo`, but `tabdo` seemed to be similar to the links in the help file for `vertical`, `horizontal`, `context`, etc.  I'd be happy to change it if I messed that portion up.

Thanks for this plugin, I really enjoy it.  :beers:


SirVer:
great little feature! Thank you.

Issue #921, opened 2017-12-22
$VISUAL placeholder is cut off by one character


ap-0:
Hello.

for the following simple snippet example:
```
snippet foo
if foo: ${1:${VISUAL:bar}}
endsnippet
```
the placeholder `bar` (or any other) will lose it's last character. Expanden it looks like
`if foo: ba`

In some situations I have snippets where I would like to keep the placeholder when I have no visual text selected. 

I am on Windows using Vim8.0 and Python 3.6.4. UltiSnips is the latest version updated today with Vundle.

I have found this commit that removes the last character for windows. But I do not know the reason
https://github.com/SirVer/ultisnips/commit/5a7ec5cea2824f443fbd9b70f67ab67fd8b47a54

SirVer:
Should be fixed in latest master.

Issue #908, opened 2017-10-30
How to avoid newline with VISUAL?


mMontu:
Is it possible to avoid including a new line when using the visual placeholder (`${VISUAL}`)?
For example, using the following snippet in a robot file, from [vim-snippets](https://github.com/honza/vim-snippets/blob/master/UltiSnips/robot.snippets#L132):

```
snippet wuks "Wait Until Keyword Succeeds"
Wait Until Keyword Succeeds    ${1:10x}    ${2:2s}    ${3:${VISUAL:Some Keyword}}
endsnippet
```

Starting with the following text,

```
        aaa
        bbb
        ccc
```

, after visually selecting the `bbb` and typing <tab>wuks I'd like to have this
  
```
        aaa
        Wait Until Keyword Succeeds    10x    2s    bbb
        ccc
```
  
```
instead of

        aaa
        Wait Until Keyword Succeeds    10x    2s    bbb

        ccc

```

andreyorst:
Fixed by #926

Issue #897, opened 2017-09-22
Expand with VISUAL loses last char in GVIM


vanabel:
It is wired that the following configuration works for vim in cygwin, but not works for gvim on windows 7.

```vim
" vim: fdm=marker fdl=0 fdls=0
" Check the operating system {{{
if !exists( 'g:os' )
  if has( 'win32' ) || has( 'win64' )
    let g:os  = 'Windows'
  elseif has( 'win32unix' )
    let g:os = 'Cygwin'
  else
    let g:os  = substitute( system( 'uname' ), '\n', '', '')
    "g:os = Windows,Cygwin,Darwin,Linux
  endif
endif
" }}}
" User's variable {{{
" $VIMHOME {{{
if g:os != 'Windows'
  let  $VIMHOME = expand( $HOME . '/.vim/' )
else
  let $VIMHOME = expand( $HOME . '/vimfiles/' )
endif
" }}}
" $USRVIMD {{{
let $USRVIMD    =expand( $VIMHOME.'myvim/' )
" }}}
" }}}
" Plugin admin {{{
call plug#begin( $USRVIMD . 'plugged' )
" for snips
Plug 'Sirver/ultisnips'
call plug#end()
" }}}
" Configuration for Ultisnips {{{
set rtp +=$USRVIMD
let g:UltiSnipsSnippetsDir = $USRVIMD. 'UltiSnips'
" }}}
" Set tex as default formart {{{
au VimEnter,BufEnter *.tex set ft=tex
" }}}
```
Sorry for the redundant code, which makes vimrc fold correctly. I also add a system detect function, which will use plugin directory correctly. The plug manager is vim-plug.

If we only concern with UltiSnips, the above vimrc install it and set the Snippets directory, in which I have a `tex.snippets` containing only one snippet:

```vim
snippet frac "Fraction" wA
\frac{${1:${VISUAL:nom}}}{${2:denom}}
endsnippet
```

I test it with a new file called `test.tex`, in which I type `frac`, idearly which will expand to `\frac{nom}{denom}`, it is the case on cygwin's vim while it expands to `\frac{no}{denom}`, i.e., `nom` losses the last char, I can't figure why? is it a bug?

asknet:
i'm facing the same issue in windows. I guess, it is something with line endings(CRLF)

asknet:
After further reasearch, this is breaking in windows due to this incomplete fix
https://github.com/SirVer/ultisnips/issues/863

[...]

PR #926, opened 2018-01-21
Fix character selection when using VISUAL on all platforms

j3soon:
This fix is for windows users using selection `inclusive`. (Problem occurred since #868 )
> The pull request #883 is fine but missed the case when using `V` in `exclusive` mode.
> I planned to fix it after it being merged. But no one merged it because of the failing tests.

Issue #897 #921 and maybe #908 all refer to this problem, so I think it should be fixed as soon as possible.

However, `.travis.yml` seems to have some problem that make all tests fail.

The error message produced when testing using Travis CI:
It'll keep trying until the time is out.

    W: The repository 'http://ppa.launchpad.net/kalakris/tmux/ubuntu trusty Release' does not have a Release file.
    W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
    E: Failed to fetch http://ppa.launchpad.net/kalakris/tmux/ubuntu/dists/trusty/main/binary-amd64/Packages  404  Not Found
    E: Failed to fetch http://ppa.launchpad.net/kalakris/tmux/ubuntu/dists/trusty/main/binary-i386/Packages  404  Not Found
    E: Some index files failed to download. They have been ignored, or old ones used instead.

@seletskiy I had tried to fix the problem occurred in Travis CI, but I didn't find any solution. I think the problem is in `.travis.yml` line 14:

    until sudo add-apt-repository ppa:kalakris/tmux -y; do sleep 10; done

### DIFF
--- a/pythonx/UltiSnips/text_objects/_visual.py
+++ b/pythonx/UltiSnips/text_objects/_visual.py
@@ -15,7 +15,6 @@ from UltiSnips import _vim
 from UltiSnips.indent_util import IndentUtil
 from UltiSnips.text_objects._transformation import TextObjectTransformation
 from UltiSnips.text_objects._base import NoneditableTextObject
-import platform
 
 _REPLACE_NON_WS = re.compile(r"[^ \t]")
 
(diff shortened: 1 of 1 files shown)


SirVer:
The fix looks correct to me. I fixed travis builds (only for py2.7 and linux though), so I rebased your work on top of the unbroken master. Let's see if this runs through.

j3soon:
@SirVer Yes, the test has finally passed! 👍

SirVer:
@j3soon Thank you very much!

PR #929, opened 2018-01-31
don't clean 1st tabstop if it's not empty

lacygoill:
Currently, in the example [Autojump from tabstop when it's empty](https://github.com/SirVer/ultisnips/tree/master/doc/examples/autojump-if-empty), the text in the 1st tabstop and the 2 parentheses around it are removed unconditionally. I think they should only be removed if the user didn't insert anything.


lacygoill:
There was another small error in one of the example given in the documentation.

PR #936, opened 2018-02-12
removed default folding

SnowCone161803:
Hi there!

Would it be possible to remove the default folding method as this is making it more difficult to use my own folding method?

I was thinking either remove it or make the default a variable that is set here.

What do you think?

### DIFF
--- a/ftplugin/snippets.vim
+++ b/ftplugin/snippets.vim
@@ -8,10 +8,6 @@ let b:did_ftplugin = 1
 let s:save_cpo = &cpo
 set cpo&vim
 
-" Fold by syntax, but open all folds by default
-setlocal foldmethod=syntax
-setlocal foldlevel=99
-
 setlocal commentstring=#%s
 
 setlocal noexpandtab


SnowCone161803:
tests passing on this branch

Travis seems to be having difficulty with a PPA...

```
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
W: The repository 'http://ppa.launchpad.net/kalakris/tmux/ubuntu trusty Release' does not have a Release file.
E: Failed to fetch http://ppa.launchpad.net/kalakris/tmux/ubuntu/dists/trusty/main/binary-amd64/Packages  404  Not Found
E: Failed to fetch http://ppa.launchpad.net/kalakris/tmux/ubuntu/dists/trusty/main/binary-i386/Packages  404  Not Found
E: Some index files failed to download. They have been ignored, or old ones used instead.
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
```

SnowCone161803:
Hmm that's a shame.

Do you know of any justification for this being The Vim Way?

SirVer:
> Do you know why that this is The Vim Way?

I can't answer that. I linked you to the official Vim documentation that states how you are supposed to overwrite settings in plugins that you do not like. I did not design Vim, so I cannot tell you how this design came to be. But it is the way Vim expects it to be done - hence the Vim Way. 

It solves your problem of overwriting the foldmethod though.

SnowCone161803:
No worries!

Thanks for the solution! It seems to be working perfectly.

SnowCone161803:
It does seem strange that you need to put the config for Vim in two places depending on how plugins are written.

Like, when you start off reading online it talks about putting everything in the `.vimrc`. But now sometimes you need to put it in the `after/ftplugin` folder.

What are your thoughts on it?

SirVer:
I do agree that this setup is not ideal, as you said defining everything in one place would be preferable. However, this is the documented and recommended way. there are only two approaches really:

1) stick to the best practices laid out in the Vim docs.

2) every plugin reinvents their own scheme of configuration. 

I am firmly in category 1. It allows people to learn the vim way and apply it everywhere, 2 requires to learn anew for every plugin.

SnowCone161803:
I've done some good learning on `ftplugin` to see how it all works and I'm happy about that. But I've realised that the problem that I've been having isn't to do with the default `foldmethod` for snippets and more to do with why I wanted it changed in the first place.

My snippets files were getting confusing because there were too many snippets in them. I'd used the `vim-plugins` as a guide for how to create snippets rather than the documentation, because it was quicker to get 
up and running quickly with.

In this plugin's `ultisnips` folder however, it wasn't using any of the folder structure described in (adding snippets)[https://github.com/SirVer/ultisnips/blob/master/doc/UltiSnips.txt#L570].  Once I found this, the problem was solved completely!

Thanks for the chat though. I wouldn't have found this unless I'd let my ideas be challenged!

PR #939, opened 2018-02-21
Do not expand brA snippets if there is already a character in the line

ig-dev:
```
snippet /f / "foo" brA
foo
endsnippet
```

Current behaviour:
Typing: `af<Space>`
Becomes: `afoo`
Expected: `af<Space>`

The b option should cause the snippet to only expand if the match starts at the beginning of the line. Together with the 'r' and 'A' options the snippet also expands if there is already one alphanumerical character typed (but not more).


ig-dev:
Oops, test fail. Let me fix this and create a new pull request.

PR #790, opened 2016-11-15
add function isExpandable(): check if expands, but do not expand

Konfekt:
A first take at https://github.com/SirVer/ultisnips/issues/786:

Add a function UltiSnips#isExpandable() (as neosnippet offers in guise of neosnippet#expandable()), useful for example for autocompletion plug-ins like delimitmate:
```vim
 nnoremap <expr> <tab> delimitMate#isEmptyPair() && UltiSnips#isExpandable() ? <del>:UltiSnips#Expand() : '<tab>'
```
This way, a snippet with delimiters such as
```vim
  snippet [
	[${1:text}](http://${2:address})
```
in markdown.snippets of vim-snippets, is expanded with no trailing ] (auto-inserted by the autocompletion plugin).

PR #957, opened 2018-04-02
document IsExpandable() to check if snippet is expandable without exp…

Konfekt:
…anding

Solves https://github.com/SirVer/ultisnips/pull/790#issuecomment-377808784


SirVer:
Beautiful! Thanks a lot.

Issue #949, opened 2018-03-29
Project status


andreyorst:
Hello, @SirVer, @seletskiy, and other contributors, and users of this project. I would like to gently ask about UltiSnips project status.

In Readme is stated that:
> UltiSnips was started in Jun 2009 by @SirVer. In Dec 2015, maintenance was handed over to @seletskiy.

But the last commit was 423f264 on Jun 30, 2017. It's almost a year, and last @seletskiy [activity in issues](https://github.com/SirVer/ultisnips/issues/878#issuecomment-322465420) is somwhere around august 2017. Both of you seem remain active on Github, so I assume that everithing is fine, wich is great. But...

There is no note on projects status, so I would like to ask, is there a new active developed fork of this, or project is abandoned by every maintainer and its destiny will remain unknown? Other snippet plugins that i've seen has less features than Ultisnips, so I'm really interested in this project health status. Current master contain some essential bugs, like #948, or #932 and others, wich I haven't encountered yet, but still, that interrupts work in some case.

SirVer:
@andreyorst In Nov 2016 I passed on all maintainership of my open source projects due to rising personal commitments (See [the post for my more successful open source project](https://wl.widelands.org/news/2016/Nov/15/sirver-passing-chieftain-baton-gunchleoc/) for details). My situation is unchanged, so I cannot commit to spending more time on UltiSnips myself right now and in the foreseeable future.

Stanislav (@seletskiy) was the most active contributor at that time and graciously accepted taking on maintainership of UltiSnips. It is a lot of work to maintain a high active project like Widelands, probably he can't keep up either any more. 

> There is no note on projects status, so I would like to ask, is there a new active developed fork of this, or project is abandoned by every maintainer and its destiny will remain unknown? 

I am still an active user of UltiSnips - I use it daily - and in its current form it is very usable, stable and feature rich. So even if there is no maintainership going forward, I think it remains a useful plugin for a long time to come. I do not think that there is a more active fork of UltiSnips, but I have not checked.

I think fixing the most annoying bugs is all that is required, that should not be a ton of work. However, I think it requires a hero to step forward with the interest of investing time into UltiSnips. I will add anybody interested in being a maintainer to the project admins, I am also open to moving UltiSnips into an organization and have its development lead by a group.

seletskiy:
[...]

PR #959, opened 2018-04-06
Update issue_template.md

andreyorst:
As I've proposed in https://github.com/SirVer/ultisnips/issues/949#issuecomment-379158188 I've added comments that will be deleted automatically, so user wont need to do this manually and resulting issue will not accidentally contain any unnecessary text.


SirVer:
The idea with the comments is great! I left a few comments for discussion and in general I find the template is now a bit long. I'll try to trim it down before merging when I have some time.

.github/issue_template.md
```diff
@@ -1,22 +1,45 @@
...
+-->
 
-- fill out the table below the line of dashes and keep it at the end of your
-  issue text. The provided answers below are examples, only keep one of them or
-  replace them entirely with something else. But please provide an answer for
-  every line.
+<!-- Issue template:
+- First please provide clear explanations of issue type. What causing it,
+  and how we could test it.
+- Provide short and understandable steps to reproduce the issue in the list
```
SirVer: not a fan of these steps list. In my experience they are never provided in a useful way.

.github/issue_template.md
```diff
@@ -1,22 +1,45 @@
...
+  below.
+- Fill out the table below ## symbols and keep it at the end of your
+  issue text. Please provide an answer for every line.
 - provide a minimal viable repro case, ideally following
```
SirVer: This is the most important line for me. If people do this, we will have a 100% reliable way of reproducing. Everything else might not have all information.

andreyorst:
Ok! I've tried to cover everything, so yes it is bit long now, but since they are deleted eventually I thought that it is not a big problem.

.github/issue_template.md
```diff
@@ -1,22 +1,45 @@
...
+**Steps to reproduce**
+
+1.
+2.
+3.
+
+
+## 
+<!-- Please fill theese fields -->
+- **Operating System**:  <!-- e.g. Windows XP / Ubuntu Linux / Mac OS 10.5 -->
```
SirVer: I think this has the danger that people will write their answer inside the comment block. That is not terrible though, since we can edit the bug they submit and change the lines.

andreyorst:
> It is unlikely that issue posters will read the full text that we write, so it is likely that they will miss information.

I got what you mean, so I'll work more on template but little later.

andreyorst:
@SirVer I think that it should be mentioned that neovim isn't supported. Somewhere at the beginning of template. So neovim related issues will popup more rarely. Your thoughts?

SirVer:
@andreyorst I gave this another go and trimmed it further down. I think it is in good shape now. If you like it like that, let's merge it.

andreyorst:
Well, thats strange, but some comments are actually not colored as comments in github editor.
image
Seems like editor bug, but because markdown specification tells that linebreak is sort of a delimiter, I think that editor thinks that comment block ended, but the resulting renered issue doesn't contain it. Bit confusing to me.

I suggest not relaing on this behaviour, because it might change, as github team is still tweaking markdown sometimes. I'd loke to add some extra comment blocks just to safety.

SirVer:
> @SirVer I think that it should be mentioned that neovim isn't supported. Somewhere at the beginning of template. So neovim related issues will popup more rarely. Your thoughts?

Let's put this into the readme instead.

> Well, thats strange, but some comments are actually not colored as comments in github editor.

GitHub's code highlighting is buggy - but the markdown implementation is not. It doesn't matter for us here, since issue reporters will not see this text syntax highlighted ever. They will only see the plain text representation. I think the plain text was clearer before e697b5c and I suggest to revert that commit.

andreyorst:
> Let's put this into the readme instead.

Good point. Maybe then it should stated in documentation too. 
> I think the plain text was clearer before e697b5c and I suggest to revert that commit.

Well it's just my excessive caution. If you're being so, i'm all for reverting.

SirVer:
> Good point. Maybe then it should stated in documentation too.

Let's try to keep information in one place only. We should link from the documentation to the readme though.

SirVer:
@andreyorst Are you happy with this PR? Can it go in?

andreyorst:
@SirVer absolutely

SirVer:
@UltiBot merge

SirVer:
@UltiBot merge

PR #962, opened 2018-04-09
Remove travis caching and avoid rate limiting. 

SirVer:
The caching only saves a few minutes and seems brittle.


SirVer:
@UltiBot merge

PR #969, opened 2018-04-11
Add note about contributing to traceback

andreyorst:
There too many issues with traceback only. While it is good to have traceback, it is sell useful when it is the only info in the issue description. I think that contributing manual should be mentioned here even if we have issue template now. At least further new issues with traceback may follow contributing guide.


SirVer:
👍 Makes sense.

SirVer:
@UltiBot merge

PR #975, opened 2018-04-15
Docs: alternative implementation of IsExpandable function

unclechu:
To deal with space characters.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1798,11 +1798,24 @@ listed in the docs.
 6. FAQ                                                        *UltiSnips-FAQ*
 
 Q: Do I have to call UltiSnips#ExpandSnippet() to check if a snippet is
-  expandable ? Is there instead an analog of neosnippet#expandable ?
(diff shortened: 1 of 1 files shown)


unclechu:
@SirVer Does it look okay?

SirVer:
@unclechu Yes, the doc update looks great to me. I am just wondering if we can test this somehow and make sure it does not regress. But I have no good idea how to test something in the documentation.... Do you have an idea?

andreyorst:
Well we can just feed keys to vim, like <kbd>:</kbd><kbd>h</kbd><kbd>ultisnips</kbd><kbd>/</kbd><kbd>function UltiSnips#IsExpandable()</kbd><kbd>V</kbd><kbd>/</kbd><kbd>endfunction</kbd><kbd>y</kbd><kbd>Ctrl</kbd><kbd>w</kbd><kbd>k</kbd><kbd>P</kbd> and we will have a buffer with copied function contents. But that's silly

SirVer:
I did not come up with a better idea for testing this, so let's get it in as is. Thanks for the contributions @unclechu !

@UltiBot merge

PR #986, opened 2018-05-22
Add au TextChangedP to make autotrigger work with completion windows open

TheZoq2:
Currently, autotrigger snippets do not work when a  completion window is open because the `TextChangedI` autocmd does not get triggered. vim8 added a new autocmd `TextChangedP` which is [now also supported](https://github.com/neovim/neovim/pull/8377) by neovim.

For reference, I originally reported this issue to the YouCompleteMe repository https://github.com/Valloric/YouCompleteMe/issues/2524 which led me to report the issue to the neovim repository https://github.com/neovim/neovim/issues/7443

This also does not work with older vim/nvim versions that do not support `TextChangedP`. What's the best way to work around this, should I check for a specific vim version or is there a way to tell if a specific autocmd is available?

If necessary I can also add a test for this.


SirVer:
It seems like this was added after 8.0 was released, i.e. is not in vim8 necessarily: https://github.com/vim/vim/blame/b1c9198afb7ff902588b45fbe44f0760a9f48375/runtime/doc/autocmd.txt#L346. This is also why the Vim 8 test fails on CI.

This means you need to check the vim version, including the patch level: https://vi.stackexchange.com/questions/2466/how-do-i-know-the-patch-version-of-my-vim

Please also add a test that you probably want to conditionally skip, an example how to do this is here: https://github.com/SirVer/ultisnips/blob/790b14dcbd015f8ed296ce2d4eca7391b6a90dc9/test/test_Editing.py#L135

chemzqm:
@TheZoq2 You can just use 
```
if exists('#TextChangedP')
endif
```
to detect.

SirVer:
Thank you!

PR #990, opened 2018-06-13
Unite doesn't work as replacement for UltiSnipsListSnippets (mapping is clobbered)

### LINKED ISSUE #212: UltiSnips Supertab integration mysteriously stopped working
Hi,

UltiSnips has mysteriously stopped working for me. Opening a new file and typing "box<Tab>" gives me the message "Pattern not found". My vimrc is [here](https://github.com/vaibhavsagar/dotfiles/blob/master/vimrc) and I haven't been able to pinpoint the issue. I've tried commenting out my UltiSnips config options to no avail. I use Python 3 on Windows and I would really appreciate some guidance. Thanks!

idbrii:
If you follow the help setup for Unite, and the binding key (F12) is the
same as UltiSnipsListSnippets, then it doesn't work because we clobber
whatever is bound to UltiSnipsListSnippets in
after/plugin/UltiSnips_after.vim. We can skip this clobbering by
defining the b:did_after_plugin_ultisnips_after flag.

I don't understand why UltiSnips_after is required to make supertab
work, but it was introduced to fix #212. As a better fix to #212, I'd
suggest checking for supertab or if supertab is bound to the expected
key sequence instead of always stomping the bindings.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1773,6 +1773,11 @@ interface with matching snippets. Pressing enter will expand the corresponding
 snippet. If only one snippet matches the text in front of the cursor will be
 expanded when you press the <F12> key.
 
+To use the same mapping as |g:UltiSnipsListSnippets|, you need to disable after
+cleanup: >
+  let b:did_after_plugin_ultisnips_after = 0
+<
+
 Supertab - UltiSnips has built-in support for Supertab. Just use a recent
 enough version of both plugins and <tab> will either expand a snippet or defer
 to Supertab for expansion.


SirVer:
Can you not bind `g:UltiSnipsListSnippets=<Nop>` ([docs](http://vimdoc.sourceforge.net/htmldoc/map.html#%3CNop%3E)) for the same effect? In the end you do not want UltiSnips to own this mapping anymore, you want Unite, right? Or maybe I am misunderstanding the problem.

idbrii:
@SirVer I was wrong. Nop is not valid. You can map something to Nop, but not Nop to something.

Using this:

    let g:UltiSnipsListSnippets        = '<Nop>'

Maps those five characters. So now if I try to make a mapping like `map  <F8>  <Nop>` in insert mode, my snippets are listed.

I think an option to disable g:UltiSnipsListSnippets would be a better option than my PR? (maybe blank to prevent the map?)

lacygoill:
FWIW, the way I deal with these kind of issues in general, is by mapping the feature to a function key which is not available.

Vim recognizes the function keys [`F1` up to `F37`](https://github.com/vim/vim/blob/8858498516108432453526f07783f14c9196e112/src/keymap.h#L194).

On my keyboard, the biggest function key is `F12`, so I can use everything beyond that as a way to disable a feature.

You can also combine a function key with the shift modifier. So, as an example, you could use one of these statements:

    let g:UltiSnipsListSnippets = '<f20>'
    let g:UltiSnipsListSnippets = '<s-f20>'

PR #997, opened 2018-07-01
add option to apply snippets automatically when only one candidate

ZSaberLv0:
When using `g:UltiSnipsListSnippets` if there is only one candidate, apply it automatically

I think it's more convenient to use, especially when you can not remember the full abbr key of a snippet

For example:

```
let g:UltiSnipsListSnippets = "<c-o>"
let g:UltiSnipsAutoApplyForSingleCandidate = 1
```

Let's say, if I have a snippet with `exists` as abbr key, and typed `exist`, I can still use `<c-o>` to apply it immediately, without the annoying `1` then `Enter` to choose the single candidate, and without concerns whether there is `s` in `exist(s)`, further more, I can even use `ex<c-o>` to "complete" the snippet

Why not use other complete engine's suggestion (such as YCM)? Because the snippet may be placed much lower of the complete popup if there are other semantic complete candidate


SirVer:
Thank you very much for taking the time writing up the feature!

Howeve, I consider this is putting a bit of a stretch on the listing feature: For example I frequently use the list to find which snippets are even defined, not necessarily trigger one snippet - auto triggering would be undesirable then. Also, why is this feature on list (vs expand)? And why does it need another option, i.e. is not always on?

All things considered, I understand the use case, but I find that the overloading of the listing feature + the extra option increase complexity and understandability significantly, which makes this unattractive for me as a maintainer. There is also a lack of engagement on this PR, which indicates that the interest in this feature is low. I hope that you understand my decision to not merge this feature in at this point in time. Thanks for taking the time to contribute!

ZSaberLv0:
That's OK, but would you please consider some other ways to achieve similar goals? for example:
* for snippet `exists`, when typed `exi` + `g:UltiSnipsExpandTrigger`
    * if only one snippet match, expand it automatically
    * if more than one match, behaves same as `g:UltiSnipsListSnippets`

SirVer:
I have thought about this before and I think it would be a good option to have. It probably needs to be behind a feature flag for folks that use <tab> for all completions.

PR #1001, opened 2018-07-17
Allow users to disable UltiSnipsListSnippets 

idbrii:
Provide a way for users to disable UltiSnipsListSnippets (when using unite or other alternative) because `<Nop>` doesn't work.

Better solution for #990.


SirVer:
Thanks! I think this is a very reasonable feature to have.

Issue #1018, opened 2018-10-07
Ultisnip not using g:UltiSnipsJumpForwardTrigger?


huangzonghao:
**Expected behavior:**
Setting `g:UltiSnipsJumpForwardTrigger` should give us the key-binding for jump forward function.

**Actual behavior:**
`g:UltiSnipsJumpForwardTrigger` is not actually being used by the plugin?

**Steps to reproduce**
Just global search the code base of the plugin and look for the keyword `UltiSnipsJumpForwardTrigger`, it seems the plugin never actually map it to anything.

-----

- **Operating System**: 
- **Vim Version**: 
- **UltiSnips Version**: 
- **Python inside Vim**: 
- **Docker repo/vimrc**:

dimbleby:
Your searching is bad.

See [here](https://github.com/SirVer/ultisnips/blob/6fdc3647f72e0a1f321ea6bd092ecd01f7c187ba/pythonx/UltiSnips/snippet_manager.py#L864) and [here](https://github.com/SirVer/ultisnips/blob/6fdc3647f72e0a1f321ea6bd092ecd01f7c187ba/pythonx/UltiSnips/snippet_manager.py#L348-L351)

Did you try setting the forward trigger and see something not work?

SirVer:
I believe this patch is incorrect. The triggers are correctly used as pointed out by @dimbleby. Closing.

PR #1019, opened 2018-10-07
Fix jump forward and jump backward mappings

huangzonghao:
A Fix for #1018


SirVer:
I believe this to be an incorrect patch, see #1018. Closing for this reason.

PR #1043, opened 2018-12-21
Add a new item 'kind = "snippets"' for completion engines to filter on.

### LINKED ISSUE #1042: Feature Request - Expand Completion Item Properties
The current creation definition of the completion items is roughly short ([Deoplete Source Example](https://github.com/SirVer/ultisnips/blob/6fdc3647f72e0a1f321ea6bd092ecd01f7c187ba/rplugin/python3/deoplete/sources/ultisnips.py#L17)). At least the property `kind` could be pretty useful but setting it to a custom value like `snippet`. I illustrate a solution at [this comment](https://github.com/SirVer/ultisnips/issues/376#issuecomment-449409513) where this property would improve the accuracy intensively. Snippet trigger are not unique in the set of all completion suggestions, so differ them by their `kind` would be awesome.

Btw: Awesome plugin :+1:

weilbith:
Fixes #1042.

Maybe the name of 'kind' which is defined manually here should be defined globally. Not sure where this is possible, since the _Deoplete_ source is written in Python.


weilbith:
Sry have forgot to adjust the test cases for this.

SirVer:
Thanks!

PR #1048, opened 2019-01-10
Give the `abort` argument to all Vim functions

lacygoill:
It prevents Vim from processing the rest of the statements inside a function, after having encountered an error.

It also gives shorter and more relevant stack traces in case of an error.


SirVer:
Thanks!

PR #1049, opened 2019-01-10
Prioritize local mappings over global ones

lacygoill:
If the user has a global mapping which begins like a local one, they probably expect the local mapping to win.
Without `<nowait>`, they would have to wait for the timeout to occur.


SirVer:
Thanks you!

PR #1050, opened 2019-01-16
Use blackhole register for select-mode deletion

simlei:
This simple PR is for having SELECT-mode mappings not interfere with the clipboard register.
Since Ultisnips basically assumes control over SELECT mode w.r.t. some mappings, imposing a life cycle. There is nothing the user can do about it. When executing a snippet with a default placeholder and whishing to `<Del>`ete it in place, the local clipboard will inevitably be overwritten with the default placeholder.

Other SELECT-mode mappings other than the three changed ones that Ultisnips may generate are in the opinion of the author also better of using the `"_` blackhole register, but those are at least with some effort changeable on user side. The three ones that were change, are not.

I would be glad if you were to accept the PR.


SirVer:
> Other SELECT-mode mappings other than the three changed ones that Ultisnips may generate are in the opinion of the author also better of using the "_ blackhole register, but those are at least with some effort changeable on user side. 

I did not find others by grepping the code base. Can you point out which other ones are generated?

Thanks for the patch. I think this is exactly the right thing to be doing.

simlei:
I did not find other `smap`s either in a quick look through my vimrc related to ultisnips -- must have confused them with some I had related to another plugin.
I'm glad you merged :) thanks!

PR #1055, opened 2019-02-08
Fix snippets breaking when opening preview window

### LINKED ISSUE #681: ultisnips forgets tabstops when preview window opens

### LINKED ISSUE #822: UltiSnips refuses to move to next tab stop after using YouCompleteMe semantic completion

thomasfaingnaert:
This fixes snippet tabstops no longer working when the preview window is opened while a snippet is active.

Fixes #681 
Fixes #822


SirVer:
Thanks!

Issue #948, opened 2018-03-26
Custom snippets doesn't work


notaduck:
It seems that my own snippets added with `UtilSnipsEdit` doesn't work, however snippets from `hpnza/vim-snippets` works just fine.

- NeoVim => `0.2.2` 
- Plugin manager => vim-plug   
  

pice of init.vim:
```
 " UtilSnips {{{

let g:UltiSnipsSnippetDirectories = ['~/.vim/UltiSnips', 'UltiSnips']

 " Trigger configuration. Do not use <tab> if you use https://github.com/Valloric/YouCompleteMe.

let g:UltiSnipsExpandTrigger="<a-tab>"
let g:UltiSnipsJumpForwardTrigger="<c-b>"
let g:UltiSnipsJumpBackwardTrigger="<c-z>"

" If you want :UltiSnipsEdit to split your window.
let g:UltiSnipsEditSplit="vertical"

	" }}}

```

```
$ ls UltiSnips/
xml.snippets
```

```
$ cat xml.snippets 
snippet tagc "description"
<OSMWayType> $1 </OSMWayType>
$0 
```

According to me it looks like everything should be okay, however if I type `tagc` in a `.xml` file and `A-tab` nothing happends.

andreyorst:
I'm expiriencing the same issue. For two weeks already.
I've installed ultisnips, created some snippets, and nothing happened. I've double checked documentation, and even vimcasts, on youtube, but still nothing. I already started to think that Ultisnips isn't workiing at all, until I've installed `hpnza/vim-snippets`, and ultisnips started to work with them, but not with my own snippets.

I'm using `neovim` from master branch with `Vundle` as a plugin manager.
![image](https://user-images.githubusercontent.com/19470159/37961753-c914863c-31c1-11e8-86a9-d166f7f1fd39.png)
I have pretty huge config for ultisnips, so I'll [post a link to it](https://github.com/andreyorst/dotfiles/blob/7393f7c209677d4b7b3ede57c2914a4e5c3cbcb1/config/nvim/plugin_configs.vim#L122), rather then copypaste it here.

This also can be reproduced with minimal `init.vim` containing only Ultisnips plugin and [theese recomendations](https://github.com/SirVer/ultisnips#quick-start)

andreyorst:
[...]

PR #1085, opened 2019-05-21
If UltiSnipsSnippetDirectories only contains a single directory, expand `~`

s-rc:
Fixes https://github.com/SirVer/ultisnips/issues/948. The issue was that there was an early termination condition that, if hit, meant the specified directories would never have ~ expansion performed, so absolute paths had to be specified in the config variables.

With this patch, ~ can be used in the config and will be parsed properly under both conditions.


SirVer:
Thank you!

PR #1106, opened 2019-10-04
Text objects (iS/aS) for snippets files

mg979:
- iS: inside snippet
- aS: around snippet (including empty lines that follow)


SirVer:
Looks good, thanks!

Issue #1057, opened 2019-02-15
Ternary replacement operator not working as exepected anymore


PlasmaHH:
**Expected behavior:**
Replacement to either A or B but never both in the ternary replacement operator. (It has been working before). That is in this particular case when $1 is empty it should display and when it contains something it should display A

**Actual behavior:**
B is always inserted, no matter the match, so we get B or AB but can't get A

**Steps to reproduce**
Use the following snippet:

```
snippet test
result $1
optional: '${1/(.+)|(.*)/(?1:A:B)/g}' full: '$1'
endsnippet
```
then do
test<tab> and play with it and observe that you can never get B to not display.

I had the more complex version of that snippet now for years and it was always working and I noticed recently that it breaks but can't tell in which version. It does so in a "clean" vim without any other plugins installed.

-----

- **Operating System**:  Linux
- **Vim Version**:  8.1.687
- **UltiSnips Version**:  6fdc3647f72e0a1f321ea6bd092ecd01f7c187ba
- **Python inside Vim**: 3.7.2
- **Docker repo/vimrc**:

SirVer:
I am able to reproduce this with my Vim setup here at home, but when I added a test case for it, the test case came back correct for the version I tested with. I created a PR to see what CI shows us here.

SirVer:
Actually, I brought this down to an inconsistency in python regex engines:

```python
import re
reg = re.compile(r"(.+)|(.*)", re.DOTALL)
replace = lambda m: print(m)
reg.subn(replace, "a", 0)
``` 

yields 
~~~

<_sre.SRE_Match object; span=(0, 1), match='a'>
('', 1)
~~~

inside the docker container where I run the tests and

~~~
<re.Match object; span=(0, 1), match='a'>
<re.Match object; span=(1, 1), match=''>
~~~

on my MacOS dev machine. Testing further, it seems to be that this behavior changed in Python 3.7 (which returns 2 matches) while all other Python versions before returned only one. Python 3.8 also returns 2.

~~~
docker run -it python:3.6-stretch /bin/bash
root@48e8dbf1be03:/# python
Python 3.6.9 (default, Oct 17 2019, 06:49:31)
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> reg = re.compile(r"(.+)|(.*)", re.DOTALL)
>>> replace = lambda m: print(m)
>>> reg.subn(replace, "a", 0)
[...]

PR #1116, opened 2019-10-30
Add test cases for #1057.

SirVer:
I can reproduce #1057 on my home Vim setup, but this test case did not fail for me in a quick test. I want to add these tests to see what CI makes of them.


SirVer:
Closing this, since we figured out that the real problem was in fact with inconsistency in python regex engines.

PR #1129, opened 2019-11-05
Remove support for Python2

SirVer:
Python 3 is > 10 years old and Python 2 will no longer be supported in a few months. Time to get rid of it, simplifying the code base (slightly) and opening the opportunities to use the Py3 quality of life improvements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.pyc
 *.swp
 doc/tags
-.ropeproject/
+/.ropeproject/
+/.mypy_cache/
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ sudo: required
 services: docker
 
 env:
-  - VIM_VERSION="7.4" PYTHON_IMAGE=2.7-stretch TAG=vim_74_py2
-  - VIM_VERSION="8.0" PYTHON_IMAGE=2.7-stretch TAG=vim_80_py2
-  - VIM_VERSION="8.1" PYTHON_IMAGE=2.7-stretch TAG=vim_81_py2
-  - VIM_VERSION="git" PYTHON_IMAGE=2.7-stretch TAG=vim_git_py2
   - VIM_VERSION="7.4" PYTHON_IMAGE=3.6-stretch TAG=vim_74_py36
   - VIM_VERSION="8.0" PYTHON_IMAGE=3.6-stretch TAG=vim_80_py36
   - VIM_VERSION="8.1" PYTHON_IMAGE=3.6-stretch TAG=vim_81_py36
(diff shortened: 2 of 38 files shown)


rubin55:
Quick quip here, say, macOS Mojave still comes with Python 2.7 as the default python + MacVim is linked against that. I suddenly was getting large swaths of errors and I think it is because of the removal of python2 support from UltiSnips a few days ago.

Long story short: UltiSnips won't work out of the box anymore on macOS, up and including current Catalina (10.15). For more info about what apple will do post 10.16, see here btw: https://www.macobserver.com/analysis/macos-catalina-deprecates-unix-scripting-languages/

While I appreciate that Python 2.x is dead as a doorknob, this change is, from the perspective of a mac user in the coming 1.5 years, not the best one ever.

SirVer:
@rubin55 I am a full time mac user and it seems reasonable to me to expect people to `brew install macvim` or similar to get access to Vim. I will no longer support Python2 going forward.

Another reasonable option is to pin to the 3.2 release of UltiSnips which still supports Python2.

patrick96:
Hi, I just wanted to report that this broke my VIM on some occasions. 

The issue was that some plugins call `has('python')` to figure out whether or not to use python. But by calling that function it loads python2 if it is available. After that ultisnips spews `This Vim cannot execute :py3 after using :python` errors. I was able to fix this by adding `call has('python3')` at the top of my `.vimrc`.

The real solution would of course be for every plugin to first call `has('python3')` before it calls `has('python')` (if it supports python3), but that is not your problem :wink: 

Just wanted to let you know and maybe help other people that stumble upon this issue.

ZSaberLv0:
mac os still comes with python2.7 as default, so I suggest at least place some info in doc or README about python2 compatibility (using old release/tag/branch, etc)

SirVer:
@ZSaberLv0 Could you send a Pull request where you would have searched for such information and how you would have liked it to be phrased?

ZSaberLv0:
I think the best solution for now, is using vim-plug like this

```
if has('python3')
    Plug 'SirVer/ultisnips'
else
    Plug 'SirVer/ultisnips', {'tag': '3.2', 'dir': get(g:, 'plug_home', '~/.vim/bundle') . '/ultisnips_py2'}
endif
```

jordwalke:
It's really unfortunate that this change was made because MacVim is distributed as a prebuilt binary and OSX still only comes with 2.7 installed. Many MacVim (maybe the largest portion) do not install MacVim through brew. I can pin to 3.2, but it's harder to recommend that other people pin to 3.2 when communicating instructions for setting up a decent Vim environment and sharing those instructions with people.

PR #1130, opened 2019-11-05
Fix deoplete source

facto:
Deoplete recently made a change that broke source base imports. See this commit in a repo by the author of deoplete that fixes the problem the same way I do here:

https://github.com/Shougo/neosnippet.vim/commit/23c97432929721af9130a163faa442dce76b1184


SirVer:
Thanks!

Issue #907, opened 2017-10-30
find_all_snippet_files and _file_to_edit search different paths


froody:
TL;DR: `_file_to_edit looks` in `$PWD` by way of `os.path.realpath()`, `find_all_snippet_files` does not look in `$PWD`, so `:UltiSnipsEdit` opens a file that is not read by `UltiSnips_Manager.expand()`

On a default install, with no setting of UltiSnips vars in my .vimrc other than:
```
let g:UltiSnipsExpandTrigger="<c-l>"
let g:UltiSnipsJumpForwardTrigger="<c-b>"
let g:UltiSnipsJumpBackwardTrigger="<c-z>"
let g:UltiSnipsEditSplit="vertical"
```
by default, I get
`g:UltiSnipsSnippetDirectories = ['UltiSnips']`

When I edit code in one of my projects, I run `:UltiSnipsEdit` which opens up a new file in `$PWD/UltiSnips/$ft.snippets`, where `$PWD` happens to be the project root, e.g. `~/src/project_name` . I create a new UltiSnip `foo` in that file and save the .snippets, switch over to the split with my project source code, type `foo<c-l>` and nothing happens.

After digging around in the code I found that _file_to_edit resolves here: https://github.com/SirVer/ultisnips/blame/master/pythonx/UltiSnips/snippet_manager.py#L749
`_get_file_to_edit` calls `find_snippet_files(ft, 'UltiSnips')`, and `find_snippet_files appends `os.path.realpath('UltiSnips')` to the list of paths, i.e. `$PWD/UltiSnips`. So far, so good.

When I actually try to expand my snippet `find_all_snippet_files` appends 'UltiSnips' to everything in `_vim.eval('&runtimepath')`:
https://github.com/SirVer/ultisnips/blob/master/pythonx/UltiSnips/snippet/source/file/ultisnips.py#L39

but runtimepath doesn't contain `$PWD`, so it doesn't find my file. Changing the above line to be:
```
check_dirs = _vim.eval('&runtimepath').split(',') + [os.getcwd(),]
```
 fixes the issue for me, but I don't suggest fixing it that way as it would just add more confusion.

It would be much better if you refactored the code so that `_file_to_edit` searched a proper subset of the paths known to find_all_snippet_files (e.g. excluding runtimepath) and not manually construct paths itself (e.g. like this: https://github.com/SirVer/ultisnips/blob/master/pythonx/UltiSnips/snippet_manager.py#L755-L778 )

[...]

Issue #711, opened 2016-06-20
`:UltiSnipsEdit` and snippet search paths have unmatching and weird semantics


xorange:
Hi fellows,

I've installed UltiSnips and it did not start up well.

I use Vundle to manage plugins and installed UltiSnips. I typed `:UltiSnipsEdit` and could get into the Edit panel, so I assumed it should work fine.

But when I type `snip` and `<tab>` with it, it did not expand to the built-in snippet format. So I'm not sure it is working correctly and guessing my own snippets won't work either.

my env (I've got restricted internet access at work, so excuse me for using hand write presenting below):
vim version: `7.4.1859`
`has python`

python version: `2.7.3`

.vimrc:

```
set nocompatible
filetype off
set rtp+=~/.vim/bundle/Vundle.vim
call vundle#begin()
Plugin 'VundleVim/Vundle.vim'
Plugin 'Lokaltog/vim-powerline'
Plugin 'SirVer/ultisnips'
call vundle#end()
filetype plugin indent on

let g:UltiSnipsUsePythonVersion = 2             " according to ultisnips/doc
let g:UltiSnipsExpandTrigger='<tab>'            " tried <c-a> and failed too
let g:UltiSnipsJumpForwardTrigger='<tab>'  " tried <c-j> and failed too
let g:UltiSnipsJumpBackwardTrigger='<s-tab>'
let g:UltiSnipsEditSplit='vertical'
```

Really appreciate some help here

Thank you!

xorange:
Also, I've tried ultisnips/test_all.py and It raised an exception and exit non-successfully. 

Would it be some problem with my local python libraries/modules?

seletskiy:
> But when I type 'snip' and '' with it, it did not expand to the built-in snippet format. So I'm not sure it is working correctly and guessing my own snippets won't work either.

I don't understand what are you exactly mean by "when I type 'snip' and ''"?

xorange:
Thanks for replying.

What I mean is:

After `:UltiSnipsEdit`, we get into the snippet edit panel. According to the screencast made by SirVer (http://www.sirver.net/blog/2012/01/08/second-episode-of-ultisnips-screencast/), we could type `snip` and expand it (with a `tab`), we could get:
```
snippet 's' "Desc..." b
endsnippet
```
But I failed to do so.

BTW, I got everything working and fine on my Macbook Air at home.

seletskiy:
[...]

PR #1138, opened 2019-11-09
Fundamentally changes how snippets are found.

### LINKED ISSUE #1120: UltiSnipsEdit! will query for an option several times
When calling `:UltiSnipsEdit!`a menu pops up showing a list of choices of detected snippet files to edit. I always choose number 4 option. Then when I press enter, the same menu is shown again and I have to choose again. This happens sometimes two times and sometimes three times

**Expected behavior:**
The menu is shown only once

**Actual behavior:**
The menu is shown several times

**Steps to reproduce**
Open a header file that has multiple options, call `:UltiSnipsEdit!`

SirVer:
This is a fairly breaking change.

- Removes `g:UltiSnipsSnippetsDir`.
- Simplifies logic around `:UltiSnipsEdit`: Without bang, it opens a snippet file in a subdirectory  of your `.vim` directory. The name of the sub directory will come from the members of `[gb]:UltiSnipsSnippetDirectories`. This will create a new snippet file and its enclosing directory if it does not yet exist (see below for details). With bang, it additionally adds all existing snippet files that are currently loaded for the current buffer.
- When the user needs to select a snippet file for editing, it shows a `*` for existing files.

Fixes #1120, #907 hopefully #711.

## Explanation of search algorithm

There are 4 combinations to be considered here: with or without bang and with `[gb]:UltiSnipsSnippetDirectories` containing a single absolute path or many relative ones.

|   | Without Bang | With Bang              |
| ------------- | ------------- |--------|
| Single absolute  | All snippet files that are existing in the single directory and match the primary filetype. If there is none `<primary_filetype>.snippets` is used.  | same as without |
| Relative Paths | All snippet files for the primary filetype that exist in the `~/.vim/<path>`s are proposed. For each directory, if there is none `<primary_filetype>.snippets` is proposed.   | like without, but also adds all snippets files that are currently used for these settings by UltiSnips. This includes every `<runtimepath>/<path>/<ft>**.snippets` for all filetypes that match the current buffer  |

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -165,7 +165,7 @@ found here: https://github.com/honza/vim-snippets
 
 The UltiSnipsEdit command opens a private snippet definition file for the
 current filetype. If no snippet file exists, a new file is created. If used as
-UltiSnipsEdit! all public snippet files are taken into account too. If
+UltiSnipsEdit! all public snippet files that exist are taken into account too. If
 multiple files match the search, the user gets to choose the file.
 
 There are several variables associated with the UltiSnipsEdit command.
@@ -180,34 +180,13 @@ g:UltiSnipsEditSplit        Defines how the edit window is opened. Possible
                             |context|        Splits the window vertically or
                                            horizontally depending on context.
 
-                                                        *g:UltiSnipsSnippetsDir*
-g:UltiSnipsSnippetsDir
-                            Defines the directory where private snippet
-                            definition files are placed in in.
-
-                            As example, if the current 'filetype' is "cpp" the
-                            :UltiSnipsEdit command looks for a file to edit in
-                            this order:
-                            1. An existing
-                               g:UltiSnipsSnippetsDir."/cpp.snippets" file
-                            2. Find a matching "cpp" snippets file in
-                               g:UltiSnipsSnippetDirectories
-                            3. Create a new
-                               g:UltiSnipsSnippetsDir."/cpp.snippets" file
-
-                            Note that directories named "snippets" are
-                            reserved for snipMate snippets and cannot be used.
-                            Also, this setting does not affect where snippets
-                            are searched for, so it is possible to change this
-                            to opening files that are then not found by
-                            UltiSnips. See |UltiSnips-how-snippets-are-loaded|
-                            for details.
-
                                                         *g:UltiSnipsSnippetDirectories*
 g:UltiSnipsSnippetDirectories
-                            Configures the directories that are searched for
-                            snippets. Do not mix up this variable with the
-                            previous one. See |UltiSnips-how-snippets-are-loaded|.
+                            An array of relative directory names OR an array
+                            with a single absolute path. Defaults to [
+                            "UltiSnips" ]. No entry in this list must be
+                            "snippets". Please read
+                            |UltiSnips-how-snippets-are-loaded| for details.
 
                                                         *g:UltiSnipsEnableSnipMate*
 g:UltiSnipsEnableSnipMate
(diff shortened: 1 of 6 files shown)


ssokolow:
> Without bang, it opens a snippet file in your `.vim` directory

I'm assuming you don't mean *directly* in `.vim`. Would you mind editing that to clarify how `:UltiSnipsEdit` chooses where to put it?

SirVer:
@ssokolow I added a table of the different cases below and referred to it. PTAL

ssokolow:
Yes, I read it several times before I commented and I'm still not confident that I understand how the statement I quoted relates to the text and table below it.

SirVer:
Tried another round of rewording. @ssokolow, if this is not better to understand, could you provide a re-formulation or some specific questions that remain unanswered for you?

ssokolow:
It's better, but the disconnect still exists in these statements.

> The name of the sub directory will come from the members of `[gb]:UltiSnipsSnippetDirectories` This will create a new snippet file and its enclosing directory if it does not yet exist (see below for details).

> All snippet files that are existing in the single directory and match the primary filetype.

> All snippet files for the primary filetype that exist in the `~/.vim/<path>`s are proposed.

> When the user needs to select a snippet file for editing, it shows a `*` for existing files.

The unclear parts are as follows:

* If I have multiple things in ``[gb]:UltiSnipsSnippetDirectories` and no matching snippet file, and I call `:UltiSnipsEdit`, what will it do? Create one in the first directory in the list? Forcibly prompt for me to choose a directory?
* If I have multple existing snippet files and I call `:UltiSnipsEdit`what will it do? Force me to pick one? Open all of them? (ie. What does "are proposed" mean in this context? Command completion? forced menu? Something else?)
* What do you mean by "When the user needs to select a snippet file for editing, it shows a `*` for existing files."? Is it going to display an `:e /path/to/directory`-style list with `*` appended/prepended to some of the names like a footnote marker? Is it going to act like Zsh's completion correction feature and interpret the first "Enter" keypress as "Change `:UltiSnipsEdit` to `:UltiSnipsEdit *` and wait for the user to confirm by pressing Enter again"? Something else?

SirVer:
Thanks for the questions! These are easy to answer.

> If I have multiple things in [gb]:UltiSnipsSnippetDirectories and no matching snippet file, and I call :UltiSnipsEdit, what will it do? Create one in the first directory in the list? Forcibly prompt for me to choose a directory?

For every entry in `[gb]:UltiSnipsSnippetDirectories` it will suggest `<ft>.snippets` as potential file. If there are multiple, it will show a menu using `:inputlist` and let the user select which file to edit.

> If I have multple existing snippet files and I call :UltiSnipsEditwhat will it do? Force me to pick one? Open all of them? (ie. What does "are proposed" mean in this context? Command completion? forced menu? Something else?)

It will show a menu and let the user select which one to edit.

> What do you mean by "When the user needs to select a snippet file for editing, it shows a * for existing files."?

If the `inputlist` is used, every entry for an existing file will have a `*` at the beginning, while every not yet existing file will not. I added a screenshot to the OP to clarify.

matthijskooijman:
I also ran into this problem (with a clean install snippets edited with `:UltiSnipsEdit` were not found) and tried this branch. Seems to work fine without customizing any directory-related configuration directives. Also the `!` version is helpful for editing other files than the default one :-)

ssokolow:
Yeah. I haven't had a chance to try this branch yet, but the main branch is insane.

I don't have anything path-related customized and it looks for the snips in `~/.vim/UltiSnips` but, if I run `:UltiSnipsEdit` on a language that has no snippet files, it creates `~/bin/UltiSnips/languagename.snippets` because I was editing `~/bin/something` and I `set autochdir`.

SirVer:
Seems like there is consensus that this is an improvement. Let's have it and iterate on master.

PR #1148, opened 2019-11-22
Allow creating snippet on BufNewFile

srydell:
Hi!

First of all I want to thank you for this great plugin! It has helped me a lot.

I'm using a small plugin [srydell/vim-skeleton](https://github.com/srydell/vim-skeleton) to create templates using UltiSnips upon entering a new file. When I last updated ultisnips, the behaviour at BufNewFile had changed, and I could no longer jump through tabstops in the template snippets. A git bisect later and I found [this commit](https://github.com/SirVer/ultisnips/commit/8ff301c651d6df806c0f9305273a6c62a693bc48) to be where the behaviour changed. This PR is the smallest change I could manage that I think solves both cases.

Thanks for your time!

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -442,7 +442,7 @@ class SnippetManager:
 
         vim_helper.command("autocmd InsertLeave * call UltiSnips#LeavingInsertMode()")
 
-        vim_helper.command("autocmd BufEnter * call UltiSnips#LeavingBuffer()")
+        vim_helper.command("autocmd BufLeave * call UltiSnips#LeavingBuffer()")
         vim_helper.command("autocmd CmdwinEnter * call UltiSnips#LeavingBuffer()")
         vim_helper.command("autocmd CmdwinLeave * call UltiSnips#LeavingBuffer()")
 
--- a/test/test_Plugin.py
+++ b/test/test_Plugin.py
@@ -22,3 +22,26 @@ class Plugin_SuperTab_SimpleTest(_VimTest):
         vim_config.append('let g:SuperTabRetainCompletionDuration = "insert"')
         vim_config.append("let g:SuperTabLongestHighlight = 1")
         vim_config.append("let g:SuperTabCrMapping = 0")
+
+
+class Plugin_VimTemplate_SimpleTest(_VimTest):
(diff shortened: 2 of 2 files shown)


SirVer:
Two comments:

1) That change was made deliberately in #1055 though, I assume. @thomasfaingnaert, you made #1055 and change that line from `Leave` to `Enter`. Do you still remember why that was useful or needed? I feel that two needs are clashing here, and toggling this back and forth does not seem a very prinicipled approach to finding a solution for both use cases.

2) Is it possible to add a test for this behavior? In light of the argumentation for 1), not having tests makes it likely that this will regress in the future.

thomasfaingnaert:
It's been a while, but IIRC, it's because `BufEnter` is run _after_ Vim switches to another buffer, whereas `BufLeave` is run _before_. Because I need to check if the current or previous buffer is the preview window, I need to use `BufEnter`, since at `BufLeave` the buffer isn't changed yet. Thus, if Vim switches from a regular buffer to the preview window, it isn't detected on `BufLeave`.

I'm not sure yet how you could fix both issues, though.

srydell:
I see. We would need tests for both use cases. @thomasfaingnaert, is the way described in #822 (found it when I followed #1055) the best way to reproduce the error related to the preview window?

I can add a test for vim-skeleton. I will try to add a test for the preview, but it will probably take more time.

thomasfaingnaert:
@srydell Personally, I stumbled on the issue when using `completeopt+=preview` with vim-lsp and the python language server. When I started completion while a snippet was active, the snippet context was torn down, because apparently Vim switched to the preview window and back. It could even be that this workaround is not even needed anymore in the latest Vim, I'm not sure. #822 does seem to be the same issue, so you can use that to repro as well.

Since floating popups have been added to Vim, I haven't found much use anymore for the preview window, but it would still be nice to support it for people that do.

srydell:
Thanks @thomasfaingnaert. I will poke around a bit when I get the time. I don't want to break anything, so I'll try to test as much as I'll have time to.

SirVer:
@srydell Any update on this?

srydell:
@SirVer I haven't had the time yet. Sorry.

SirVer:
@srydell Closing this, let me know if you start working on it again. More than happy to reopen then.

Issue #1166, opened 2020-01-18
[Feature-request] Implement choices


hikerpig:
Would be useful if this kind of syntax and interaction is supported. 

Found it in TextMate manual, `${«int»|«choice 1»,…,«choice n»|}` .

For example:

```vim-snippet
color: ${1|red,green,blue|}
```

And a demo in VSCode:

![2020-01-18 22 54 57](https://user-images.githubusercontent.com/2259688/72665677-a85c5800-3a45-11ea-9e35-5cb44df0b3a5.gif)

lacygoill:
Try this:

    snippet cc "color property" bm
    color: $1`!p snip.rv = complete(t[1], ['red', 'green', 'blue'])`
    $0
    endsnippet

For more info, see `:h UltiSnips-python` and/or watch [this video](https://www.youtube.com/watch?v=JJQYwt6Diro&t=200).

---

If you have other usage questions, visit [vi.stackexchange](https://vi.stackexchange.com), and tag your question with [plugin-ultisnips](https://vi.stackexchange.com/questions/tagged/plugin-ultisnips).

SirVer:
I thought about supporting this syntax before. While @lacygoill implementation actually gives you the exact feature, it seems a common enough use case to me that might make sense to actually support it inside of the engine. The proposed syntax also does not clash with any already existing one, so it should be rather easy to actually pull this in.

What do you think, @lacygoill ?

lacygoill:
I think it would be a worthy addition, because the current way of implementing this feature can lead to an unexpected motion of the cursor (when you nest a tabstop inside another one).

Consider this snippet:

    snippet test ""
    foo ${1:bar$2`!p snip.rv = complete(t[2], ['a', 'b', 'c'])` baz}
    endsnippet

Insert `test`, then press the key to expand the snippet; you get this:

    foo bar[a | b | c] baz
        ^^^^^^^^^^^^^^^^^^
        selected

Now press the key to jump to the second tabstop; the cursor jumps right before the opening square bracket:

    foo bar|[a | b | c] baz
           ^

Finally, press `a` to select the `a` match in the list; you get this:

    foo bara baz

Which is correct and expected; what is not expected is the cursor jumping 3 characters forward.
It went from here:

    foo bara| baz
            ^

To there:

    foo bara ba|z
               ^

[...]

PR #1172, opened 2020-02-01
Feature: implement choices

hikerpig:
A sketchy implementation of Choices , releated #1166

### DIFF
--- a/pythonx/UltiSnips/snippet/parsing/base.py
+++ b/pythonx/UltiSnips/snippet/parsing/base.py
@@ -59,6 +59,7 @@ def tokenize_snippet_text(
                 if klass is not None:
                     text_object = klass(parent, token)
 
+                    # TabStop has some subclasses (e.g. Choices)
                     if isinstance(text_object, TabStop):
                         seen_ts[text_object.number] = text_object
 
--- a/pythonx/UltiSnips/snippet/parsing/base.py
+++ b/pythonx/UltiSnips/snippet/parsing/base.py
@@ -57,7 +57,10 @@ def tokenize_snippet_text(
             else:
                 klass = token_to_textobject.get(token.__class__, None)
                 if klass is not None:
-                    klass(parent, token)
+                    text_object = klass(parent, token)
+
+                    if isinstance(text_object, TabStop):
+                        seen_ts[text_object.number] = text_object
 
     _do_parse(snippet_instance, text, allowed_tokens_in_text)
     return all_tokens, seen_ts
--- a/pythonx/UltiSnips/snippet/parsing/lexer.py
+++ b/pythonx/UltiSnips/snippet/parsing/lexer.py
@@ -271,6 +271,11 @@ class ChoicesToken(Token):
 
         self.number = _parse_number(stream)
 
+        if self.number is 0:
+            raise RuntimeError(
+                "Choices selection is not supported on $0"
+            )
+
         next(stream)  # |
 
         choices_text = _parse_till_unescaped_char(stream, "|")[0]
--- a/pythonx/UltiSnips/snippet/parsing/lexer.py
+++ b/pythonx/UltiSnips/snippet/parsing/lexer.py
@@ -272,7 +272,7 @@ class ChoicesToken(Token):
 
         choices_text = _parse_till_unescaped_char(stream, "|")[0]
         self.choice_list = choices_text.split(",")
-        self.initial_text = choices_text
+        self.initial_text = "|{0}|".format(choices_text)
 
         _parse_till_closing_brace(stream)
 
--- a/pythonx/UltiSnips/snippet/parsing/lexer.py
+++ b/pythonx/UltiSnips/snippet/parsing/lexer.py
@@ -250,6 +250,41 @@ class MirrorToken(Token):
         return "MirrorToken(%r,%r,%r)" % (self.start, self.end, self.number)
 
 
+class ChoicesToken(Token):
+
+    """${1|o1,o2,o3|}"""
+
+    CHECK = re.compile(r"^\${\d+\|")
+
+    @classmethod
+    def starts_here(cls, stream):
+        """Returns true if this token starts at the current position in
+        'stream'."""
+        return cls.CHECK.match(stream.peek(10)) is not None
+
+    def _parse(self, stream, indent):
+        next(stream)  # $
+        next(stream)  # {
+
+        self.number = _parse_number(stream)
+
+        next(stream)  # |
+
+        choices_text = _parse_till_unescaped_char(stream, "|")[0]
+        self.choice_list = choices_text.split(",")
+        self.initial_text = choices_text
+
+        _parse_till_closing_brace(stream)
+
+    def __repr__(self):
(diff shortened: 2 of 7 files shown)


SirVer:
Sorry for the slow response. I just came back from vacation and my whole family is down with the flu currently, so I have quite limited time. I will definitely review this as next action on UltiSnips!

pythonx/UltiSnips/snippet/parsing/base.py
```diff
@@ -57,7 +57,10 @@ def _do_parse(parent, text, allowed_tokens):
             else:
                 klass = token_to_textobject.get(token.__class__, None)
                 if klass is not None:
-                    klass(parent, token)
+                    text_object = klass(parent, token)
+
+                    if isinstance(text_object, TabStop):
```
SirVer: Add a comment that choices token actually materialize into a TabStop instance, please.

pythonx/UltiSnips/snippet/parsing/lexer.py
```diff
@@ -250,6 +250,41 @@ def __repr__(self):
         return "MirrorToken(%r,%r,%r)" % (self.start, self.end, self.number)
 
 
+class ChoicesToken(Token):
+
+    """${1|o1,o2,o3|}"""
```
SirVer: Worthwile adding in this comment that choices are taken literally, i.e. no recursive parsing is taking place like it is for other tabstops.

pythonx/UltiSnips/text_objects/base.py
```diff
@@ -143,6 +143,7 @@ def overwrite(self, buf, gtext):
         if self.current_text == gtext:
             return
         old_end = self._end
+
```
SirVer: Revert this file?

pythonx/UltiSnips/text_objects/choices.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+"""Choices are enumeration values you can choose, by selecting index number,
+it is also a special TabStop
+"""
```
SirVer: empty line after dco comment.

pythonx/UltiSnips/text_objects/choices.py
```diff
@@ -0,0 +1,133 @@
...
+from UltiSnips import vim_helper
+from UltiSnips.position import Position
+from UltiSnips.text_objects.tabstop import TabStop
+
```
SirVer: all UltiSnips import in the same block, please.

pythonx/UltiSnips/text_objects/choices.py
```diff
@@ -0,0 +1,133 @@
...
+        # trying to get what user inputted in current buffer
+        if ctype == "I":
+            self._input_chars.append(cmd_text)
+        elif ctype == "D":
+            line_text = vim_helper.buf[cursor_line - 1]
+            self._input_chars = list(line_text[self._start.col: col])
+
+        inputted_text = "".join(self._input_chars)
+
+        if not len(self._input_chars):
```
SirVer: equivalent: `if not self._input_chars`

pythonx/UltiSnips/text_objects/choices.py
```diff
@@ -0,0 +1,133 @@
...
+                num = int(inputted_text)
+                remained_choice_list = list(self._choice_list)[num - 1: num]
+            else:
+                should_continue_input = True
+        else:
+            remained_choice_list = []
+
+        if should_continue_input:
+            # will wait for further input
+            # maybe we can inform user about this?
```
SirVer: remove? We have no current plans of figuring out the UI for that.

test/test_Choices.py
```diff
@@ -0,0 +1,48 @@
...
+    wanted = "1 is a char"
+
+
+class Choices_EmptyChoiceWillBeDiscarded(_VimTest):
+    snippets = ("test", "${1|a,,c|}")
+    keys = "test" + EX
+    wanted = "1.a|2.c"
+
+
+class Choices_WillNotExpand_If_ChoiceListIsEmpty(_VimTest):
```
SirVer: Add a few more tests for the following cases:

- Mirror:`${1|foo,blah|} - $1`
- Non-latin characters: `${1|öääö|trna|some korean?|}`
- Choices as zero tabstop: `${0|foo,blah|}
- Choices as nested element: `${1:${2|foo,blah|}}` - test both overwriting the default text as well as tabbing through to it.

SirVer:
@hikerpig Sorry for the long turnaround, but unfortunately I also caught the tail end of my families sickness and was out for a few days. I did a first pass over the code and I think in general it looks clean, great job! Let me know your thoughts on my comments.

hikerpig:
1. I tend to keep the syntax with precedent implementations, AFAIK, TextMate and vscode, the second one I use most of the time.
2. Your comments are valuable, I will fix the style issues and add more explications in comments.
3. I will try to add some documentation, but I'm very new at developing vim plugin and not very good at writing, this might take some time 🤔  

Will start working on these on the weekend.

test/test_Choices.py
hikerpig: I have a question and a problem with these tests.

* Choices as zero tabstop: `${0|foo,blah|}`, what is the expected behavior ? In my test,   `_update()` is called, but `_do_edit()` is not.

  1. So with input `test<tab>`, the result is `1.foo|2.blah`.
  2. So with input `test<tab>1`, the result is `1` .

  If that is correct , should the test named `Choices_AsZeroTabstop_SelectionWillNotWork` ? Emm, looks like a bug.
  
  
* Mirror: `${1|foo,blah|} - $1` , with input `test<tab>1`, the result is `foo - `. Here I see there is indeed a bug, I will need some time to fix this.

SirVer:
1. okay, sounds good to me
2. :)
3. If you feel uncomfortable with this, I can also take a stab at documentation. Just let me know.

test/test_Choices.py
SirVer: a) Yes, that is a nasty corner case I have not considered. UltiSnips exits a snippet as soon as `$0` is reached and will not consider any further transformations inside the snippet. In our case this is not expected behavior, because users will expect that the last placeholder still lets them select.... Since the implementation is fairly difficult, I suggest to just state in the documentation that selections are not supported on `$0` and to throw an error if people try it (add a test for the thrown error, please).

b) cool, great that we found a bug through a test i proposed :).

hikerpig:
@SirVer  Sorry for the slow response, now I'm back from the holiday so I basically only have time in the weekend.

I've updated the code, including:

1. Add those tests you mentioned.
2. Tweak the text object implementation to fix the Mirror bug.
3. Throw an error when using choices with `$0`.
4. Fix code style issues, also add some comments.

Unfortunately I'm such a bad writer, so I need your help with the documentation, please 🆘.

SirVer:
Thanks for the updated code! ❤️ 

I am well loaded right now, but I'll review soon and help with the documentation.

SirVer:
I am happy with the code. Let's get this in and I'll add documentation for it once I get around to it.

Thanks for your contribution!

PR #1183, opened 2020-02-28
autotrigger: trigger only on new characters insert

### LINKED ISSUE #1141: Autotrigger `[` then backspacing inserts stray characters
Hi everyone,

Let me start by saying that this is a pretty awesome vim plugin, it really does make writing with latex a breeze.

There is one little thing that has been bugging me lately, I use ultisnips to insert matching braces in neovim.
Until now, I had this line under ~/.config/nvim/ultisnips/tex.snippets:
~~~
snippet [ "Match square braces" A
[ $1] $0
endsnippet
~~~
.

Notice the space in front of the $1, if I don't ad it, I can't tab through it, since it only ads more braces.
This basically allows me to automatically create matching braces and then tab through them, which is really convenient.

However, there are two issues that come with this.
The first one is that, if I do not insert anything into the braces, it simply ads another ] in front of the cursor when pressing backspace, here is how:
[ ||] =========> (press tab) [ ||]],
where || represents the cursor.
If I press backspace multiple times, it simply ads more braces.

Here is another scenario in which this occurs, if I have two matching braces containing text, like:
[ This is a phrase ], then place my cursor into them and do 'ci[ ', it also inserts an extra closing brace like the example above.
It does the same, when I substitute (press s in normal mode) right in front of the opening brace.

And finally this happens with any type of brace wether it is a square one, a curly one or the normal ones.
I guess this has to do with ultisnips checking for the previously inserted character when it goes back to insert mode from normal mode, but I really don't know this. 

I know this is somewhat more of a feature request than a bug report, but I found it weird and was wondering if there were any easy workarounds for this.

This happens on all  my machines wether it is my gentoo system or my macos system, so I don't believe that the (neo)vim version is relevant, if it somehow is, it is 0.4.2 for my neovim.

I'm really not sure about my ultisnips version, but it should be the latest one, e.g. 3.1, I just did :PlugUpdate a few minutes before posting this.

My python version inside vim is 2.7.16 and 3.6.9.

I've tried this with no .vimrc (or init.vim in my case) at all, so I dare say it isn't necessary to provide one.
If someone knows a workaround for this, please let me know, it would be greatly appreciated.

### LINKED ISSUE #1157: Have Autotrigger trigger only execute by typing in insert mode
To me, the expected behaviour of the snippet
```
snippet ( "parentheses" iA
(${1:${VISUAL}})$0
endsnippet
```
would be to create a closing parenthesis only when _inserting_ an opening parenthesis `(`  (such a  snippet is very useful as it pairs parentheses and allows you to jump quickly out of them). 

However, with the snippet as defined, if I type `(a` (producing `(a|)` where `|` is the cursor position) then realise the `a` was a mistake, backspacing to delete the `a` results in the snippet triggering a second time, leaving the awkward `((|)` (instinctively one then backspaces again in an attempt to remove the extraneous `(`, and the whole situation escalates!).

Similarly, the snippet is also triggered if from normal mode I use substitute (default binding `s`) to replace the `a`.

Firstly we should determine whether it is the intended behaviour to have the `A` option produce snippets which can repeatedly fire in this way (as I said, I would expect not, but others may have a different view on this). If it is intended behaviour, perhaps this could be tagged as a feature request: Automatic snippets which trigger on insert only (or even, trigger a maximum of on time only).

seletskiy:
**This is draft pull request.**

I'm going to track related changes here and if it will work consistently in my setup for a while then it can be reviewed and merged.

Fixes #1157, Fixes #1141.

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -890,8 +890,8 @@ class SnippetManager:
 
                 if (
                     before
+                    and self._last_change[0] != ''
                     and before[-1] == self._last_change[0]
-                    or self._last_change[1] != vim_helper.buf.cursor
                 ):
                     self._try_expand(autotrigger_only=True)
         finally:


SirVer:
@seletskiy Welcome back :). Let me know when you feel this is ready for merging and I'll review then.

SirVer:
@seletskiy Is this ready now?

seletskiy:
@SirVer: sorry for the late reply. Yep, it seems to work in my setup.

seletskiy:
@SirVer: I think you can merge it.

SirVer (approved):

PR #1195, opened 2020-03-13
[vim_helper] Properly support XDG_CONFIG_HOME for vim

Foxboron:
Currently we only care about XDG_CONFIG_HOME if nvim is installed. But
vim can use this directory as well, so we explicitly want to have it on
our discovery path.

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/pythonx/UltiSnips/vim_helper.py
+++ b/pythonx/UltiSnips/vim_helper.py
@@ -218,9 +218,14 @@ def get_dot_vim():
     candidates = []
     if platform.system() == "Windows":
         candidates.append(os.path.join(home, "vimfiles"))
+
     if vim.eval("has('nvim')") == "1":
         xdg_home_config = vim.eval("$XDG_CONFIG_HOME") or os.path.join(home, ".config")
         candidates.append(os.path.join(xdg_home_config, "nvim"))
+
+    if vim.eval("$XDG_CONFIG_HOME"):
+        xdg_home_config = vim.eval("$XDG_CONFIG_HOME") or os.path.join(home, ".config")
+        candidates.append(os.path.join(xdg_home_config, "vim"))
 
     candidates.append(os.path.join(home, ".vim"))
 


pythonx/UltiSnips/vim_helper.py
```diff
@@ -218,10 +218,15 @@ def get_dot_vim():
     candidates = []
     if platform.system() == "Windows":
         candidates.append(os.path.join(home, "vimfiles"))
+
     if vim.eval("has('nvim')") == "1":
```
SirVer:
```suggestion
    if vim.eval("has('nvim')") == "1" or vim.eval("$XDG_CONFIG_HOME"):
```

and delete the rest of the code?

SirVer:
@Foxboron Can you reference where it is documented that Vim is supporting this now? Because searching in the main repository: https://github.com/vim/vim/search?q=XDG_CONFIG_HOME&unscoped_q=XDG_CONFIG_HOME  yields no indication that it does.

Foxboron:
It doesn't. People hack this together with `export VIMINIT=":source $XDG_CONFIG_HOME/vim/vimrc"`.

pythonx/UltiSnips/vim_helper.py
Foxboron: We would need to append both paths in that case, but I realized this is the case when having nvim with the current code.

```
    if vim.eval("has('nvim')") == "1"  or vim.eval("$XDG_CONFIG_HOME"):
        xdg_home_config = vim.eval("$XDG_CONFIG_HOME") or os.path.join(home, ".config")
        candidates.append(os.path.join(xdg_home_config, "nvim"))
        candidates.append(os.path.join(xdg_home_config, "vim"))
```

SirVer:
@Foxboron A  supported way by UltiSnips of achieving this seems to be to use the `MYVIMRC` variable. UltiSnips will actually correctly handle that one. Is that not working in your use case?

Foxboron:
Ah, that seems to work fine actually. I missed the 4 lines further down the candidate path detection there.

SirVer:
Nice, I am going to close this PR then.

Issue #1201, opened 2020-04-03
UltiSnipEdit command does not work


Qiguang:
**Expected behavior: UltiSnipEdit command from vim command mode will open a snippet file**

**Actual behavior: error prompt:**

> Error detected while processing function UltiSnips#Edit:
> line    6:
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/home/robin/.vim/plugged/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 851, in _file_to_edit
>     _get_potential_snippet_filenames_to_edit(snippet_dir, filetypes)
>   File "/home/robin/.vim/plugged/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 94, in _get_potential_snippet_filenames_to_edit
>     ft_snippets_files = find_snippet_files(ft, snippet_dir)
>   File "/home/robin/.vim/plugged/ultisnips/pythonx/UltiSnips/snippet/source/file/ulti_snips.py", line 26, in find_snippet_files
>     directory = os.path.expanduser(directory)
>   File "/usr/lib/python3.5/posixpath.py", line 228, in expanduser
>     if not path.startswith(tilde):
> AttributeError: 'PosixPath' object has no attribute 'startswith'

**Steps to reproduce**

1.the vimrc that I use can be found [here](https://github.com/Qiguang/.vim/blob/master/vimrc)
2.install ultisnip by vim-plug command :PlugInstall
3.open a *.c file in vim then type :UltiSnipsEdit<CR>

-----

- **Operating System**: Windows 10 WSL Ubuntu-16.04
- **Vim Version**: VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Jun 07 2019 15:35:43)
- **UltiSnips Version**: ee31ea1c7d08ced8577120b4c1d55c5a05351a75
- **Python inside Vim**: no python, python3 version:3.5.2 (default, Oct  8 2019, 13:06:37)
[GCC 5.4.0 20160609]

- **Docker repo/vimrc**:

SirVer:
UltiSnips required python >= 3.6, 3.5 is not supported.

Issue #1190, opened 2020-03-08
UltiSnipsEdit will not start


Msrgit:
**Expected behavior:**
UltiSnipsEdit and UltiSnipsEdit! should start

**Actual behavior:**129d33fc34a72ced857f3cfb72ff10caf10f0020
UltiSnipsEdit does not start. This traceback produced:

Error detected while processing function UltiSnips#Edit:
line    6:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/mark/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 851, in _file_to_edit
    _get_potential_snippet_filenames_to_edit(snippet_dir, filetypes)
  File "/home/mark/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet_manager.py", line 94, in _get_potential_snippet_filenames_to_edit
    ft_snippets_files = find_snippet_files(ft, snippet_dir)
  File "/home/mark/.vim/bundle/ultisnips/pythonx/UltiSnips/snippet/source/file/ulti_snips.py", line 26, in find_snippet_files
    directory = os.path.expanduser(directory)
  File "/usr/lib/python3.5/posixpath.py", line 228, in expanduser
    if not path.startswith(tilde):
AttributeError: 'PosixPath' object has no attribute 'startswith'

**Steps to reproduce**

.vimrc directives only:

"
"UltiSnips
"
let g:UltiSnipsUsePythonVersion=3
let g:UltiSnipsExpandTrigger="<tab>"
let g:UltiSnipsJumpForwardTrigger="<tab>"
let g:UltiSnipsJumpBackwardTrigger="<s-tab>"

1. :UltiSnipEdit
2.
3.

-----

- **Linux Mint 18.3**: 
- **VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Feb 27 2020 23:43:58)
Included patches: 1-326:**: 
- **UltiSnips Version**: 
129d33fc34a72ced857f3cfb72ff10caf10f0020
- **Python inside Vim**:
Python 3.5.2 (default, Oct  8 2019, 13:06:37) 
- **Docker repo/vimrc**:

SirVer:
Python 3.5 is not supported since it is very old. Python 3.6 is the first to be supported. Time to upgrade?

PR #1208, opened 2020-04-15
Workaround for broken pre-3.6-pathlib

mbrunnen:
I think it is a good idea to re-establish the compatibility to
python 3.5, since it is widely used, e.g. in debian stretch.

`pathlib` in python 3.5 is missing some methods, like `startswith`.
In order to avoid using pathlib, the paths are cast to `str`.

See also here: https://bugs.python.org/issue27493

referencing #1201, #1190

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: required
 services: docker
 
 env:
+  - VIM_VERSION="7.4" PYTHON_IMAGE=3.5-stretch TAG=vim_74_py35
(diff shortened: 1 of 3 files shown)


SirVer:
Stretch is supported till June 2022, so I agree that this makes sense.

However, this should also enable testing for py 3.5 in CI and update the docs of supported versions

SirVer:
Thank you @mbrunnen !!

PR #1229, opened 2020-05-17
fix to JumpForwardTrigger works when snippets are used in two or more buffers

### LINKED ISSUE #1187: JumpForwardTrigger stops working if snippets are used in two or more buffers
image

**Expected behavior:**
Pressing the key for JumpForwardTrigger jumps to next variable

**Actual behavior:**
The value of **g:UltiSnipsJumpForwardTrigger** is inserted

**Steps to reproduce**

1. Use the recommended docker container: https://github.com/SirVer/ultisnips/blob/master/CONTRIBUTING.md#reproducing-bugs
2. Add variables to the example snippet `/root/.vim/UltiSnips/all.snippets`
```
snippet example "Repro case example"
This $1 is $2 a $3 simple $4 snippet $5 example
endsnippet
```
3. start `vim`
4. Expand the snippet by typing `example` and pressing `<tab>`
5. Fill the variables and jump to the next one by typing `2`
6. Open new buffer: `:new<cr>`
7. Repeat steps 4. and 5. in the new buffer. Now the `2` gets inserted literally

-----

- **Operating System**: Recommended Docker container
- **Vim Version**: VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Mar  6 2020 16:11:23) Included patches: 1-586
- **UltiSnips Version**: e910b4fb9e276d18ed140fc492f30967ef9ace59
- **Python inside Vim**: 3.6.10
- **Docker repo/vimrc**: https://github.com/pjio/ultisnips/blob/repro-branch/docker/snippets/all.snippets

**Workaround**: In `ultisnips/pythonx/UltiSnips/snippet_manager.py` go to `def _setup_inner_state(self)` and remove the lines:
```
        if self._inner_state_up:
            return
```
I assume, the flag `_inner_state_up` doesn't get set to false which then causes this problem.

musou1500:
This PR closes #1187 .
if user switches buffer while filling placeholder, exception is throwd and `_teardown_inner_state` does not set `False` to  `_inner_state_up`. so `_setup_inner_state` does not configure keyboard mappings.
I moved  `_inner_state_up = False` statement to `finally` block in `_teardown_inner_state`.
It guarantees `_inner_state_up` is set to `False` even if throws while processing tear down.

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -481,12 +481,13 @@ class SnippetManager:
             vim_helper.command("augroup UltiSnips")
             vim_helper.command("autocmd!")
             vim_helper.command("augroup END")
-            self._inner_state_up = False
         except vim_helper.error:
             # This happens when a preview window was opened. This issues
             # CursorMoved, but not BufLeave. We have no way to unmap, until we
             # are back in our buffer
             pass
+        finally:
+            self._inner_state_up = False
 
     @err_to_scratch_buffer.wrap
     def _save_last_visual_selection(self):


SirVer:
Awesome! Thanks so much.

PR #1240, opened 2020-06-04
Python 3.5 Compatibility: Fix crash when snippet for current file does not exist 

mbrunnen:
I found another case, where ultisnips crashed. Again it is because of
the incompatibility between pathlib and os.path in python v3.5.

The problem came, when there is no snippet file for the current file
type, cf. `snippet_manager.py`, line 98.

Instead of casting every `os.path` argument to str, I just avoided to
cast to `pathlib.Path` in the first place.

Also I added some type hints.


SirVer:
No harm in adding this in, but I believe support for 3.5 could regress at any point in time, since it is not tested. Thanks!

PR #1261, opened 2020-07-29
Optimize choices, support space as a terminator for selection

hikerpig:
Should fix the bug where you can't select the first one, when there are more than 9 candidates. 
Related [issue 1166 comment here](https://github.com/SirVer/ultisnips/issues/1166#issuecomment-662881641).


SirVer:
@hikerpig Classy submission, thanks for your work! ❤️

PR #1263, opened 2020-08-04
Adds function to test if cursor is in front of a full snippet

Gordin:
This adds a `UltiSnips#ExpandableExact()` function that can be used to determine if the cursor is in front of a FULL snippet.
I'm using this to create the following mapping:
```viml
inoremap <silent><expr> <CR>
      \ UltiSnips#ExpandableExact() ? "<C-R>=UltiSnips#ExpandSnippet()<CR>":
      \ "\<CR>"
```
or
```viml
inoremap <silent><expr> <CR>
      \ (pumvisible() && UltiSnips#ExpandableExact()) ? "<C-R>=UltiSnips#ExpandSnippet()<CR>":
      \ "\<CR>"
```
This allows me to use `<TAB>`+`<S-TAB>` to cycle through results from my intellisense and use `<CR>` to expand a snippet when I press `Enter`, but only when I have the cursor on a complete snippet (and if the completion menu is open for the second one).

This is different from the answer to the FAQ question about "analog of neosnippet#expandable" in the documentation. The suggested `!empty(UltiSnips#SnippetsInCurrentScope())` in the answer will return true if the word under the cursor is the beginning of any snippet, but I don't want `<CR>` to expand a snippet that is only partially typed in.

I tried a couple different mappings but I wasn't able to create the mapping I wanted to without the new function, is there a way to get the desired mapping without the new code?

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -87,11 +87,16 @@ endfunction
 function! UltiSnips#SnippetsInCurrentScope(...) abort
     let g:current_ulti_dict = {}
     let all = get(a:, 1, 0)
+    let exact = get(a:, 2, 1)
     if all
       let g:current_ulti_dict_info = {}
     endif
-    py3 UltiSnips_Manager.snippets_in_current_scope(int(vim.eval("all")))
+    py3 UltiSnips_Manager.snippets_in_current_scope(int(vim.eval("all")), int(vim.eval("exact")))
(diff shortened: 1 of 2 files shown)


SirVer:
@Gordin Thanks for this feature! Sorry it took me so long to take a look. Would you mind adding documentation to this, then it is good to go.

SirVer:
Superseeded by #1314

Issue #1268, opened 2020-09-06
Support character escape in choices item


hikerpig:
More specifically, escape comma - which is the separator of choices but also very common as function parameter separator - by `\,` .

An example:

```
snippet q
${1:result} = nn.${2|fun1(,fun2(param1\, ,fun3(param1\, param2\, |}prarm_end)
endsnippet
```

Mentioned by @roachsinai in [a comment in issue 1166](https://github.com/SirVer/ultisnips/issues/1166#issuecomment-662567427).

hikerpig:
@roachsinai You can try this out.

roachsinai:
@hikerpig Works well. Thank you very much!

SirVer:
Thanks for this feature!

PR #1271, opened 2020-09-13
Add support to escape comma inside choice item by "\,"

hikerpig:
Related #1268 .


SirVer:
Thanks for this, really awesome!

PR #1279, opened 2020-10-05
Fix: DeprecationWarning from unescaped regex

farisachugthai:
Opening vim immediately fires a DeprecationWarning if python3.8 is used
and this behavior can be stopped by simply prefixing the string with `r`


SirVer:
Thanks so much!

PR #1287, opened 2020-10-23
Use first specified directory.

zhoosch:
Hello,

i use UltiSnips on different platforms with mostly the same directories (WSL, Win10, VM, Linux) if possible i use the same folder for different instances of VIM and Neovim. So the following minimal change will be a huge benefit for me and hopefully for others too.

**Expected behavior:**
`let g:UltiSnipsSnippetDirectories = ['one', 'two', 'UltiSnips']` 
Should affect the usage of `:UltiSnipsEdit` in a way, that always the first path specified by a user is chosen.

**Actual behavior:**
In case of more than one entries in `g:UltiSnipsSnippetDirectories` the default path in dot-dir will be used. The code which is affected here will be found in `snippet_manager.py` in function `_file_to_edit`.

I hope this will not affect anyone negative. Maybe i should add a switch to prevent changes in the expected behaviour for long term users?

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -180,6 +180,11 @@ g:UltiSnipsEditSplit        Defines how the edit window is opened. Possible
                             |context|        Splits the window vertically or
                                            horizontally depending on context.
 
+                                                        *g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit*
+g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit
+                            A single directory or absolute path name the
+                            location to edit and add UltiSnips.
+
                                                         *g:UltiSnipsSnippetDirectories*
 g:UltiSnipsSnippetDirectories
                             An array of relative directory names OR an array
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -165,8 +165,8 @@ found here: https://github.com/honza/vim-snippets
 
 The UltiSnipsEdit command opens a private snippet definition file for the
 current filetype. If no snippet file exists, a new file is created. If used as
-UltiSnipsEdit! all public snippet files that exist are taken into account too. If
-multiple files match the search, the user gets to choose the file.
+UltiSnipsEdit! all public snippet files that exist are taken into account too.
+If multiple files match the search, the user gets to choose the file.
 
 There are several variables associated with the UltiSnipsEdit command.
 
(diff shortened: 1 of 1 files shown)


SirVer:
@zhoosch I do not understand the desired behaviour? Can you take me through an example?

zhoosch:
@SirVer I'll do with an example.

I am an extensive vim/neovim so i have for example on work an Win10 vim & nvim and also a WSL vim & nvim installation. Which should use the same folder for UltiSnips.

``` vim
let g:UltiSnipsEditSplit = 'context'
nnoremap <leader>ns :UltiSnipsEdit<CR>
let g:UltiSnipsExpandTrigger = "<TAB>"
let g:UltiSnipsListSnippets = "<M-TAB>"
let g:UltiSnipsJumpForwardTrigger = "<C-j>"
let g:UltiSnipsJumpBackwardTrigger = "<C-k>"

if has('win32')
    let g:UltiSnipsSnippetDirectories=['d:\repositories\anywhere\configs\vim\snippets\UltiSnips', 'UltiSnips']
    let g:neosnippet#snippets_directory='d:\repositories\anywhere\configs\vim\snippets'
else
    let g:UltiSnipsSnippetDirectories=['/mnt/d/repositories/anywhere/configs/vim/snippets/UltiSnips', 'UltiSnips']
    let g:neosnippet#snippets_directory='/mnt/d/repositories/anywhere/configs/vim/snippets'
endif
``` 

As you see I use `:UltiSnipsEdit` for each of the systems one config for each installation. The benefit is i have the same location for snippets all over my local systems (VM's and Docker). 

Also i have the problem with the previous version (actual master branch), that i cannot only write:
`let g:UltiSnipsSnippetDirectories= '/mnt/d/repositories/anywhere/configs/vim/snippets/UltiSnips']`
because this hides `vim-snippets` snippets in my environment.

SirVer:
I understand your use case. We used to have a separate Variable for that, I think it was called `g:UltiSnipsSnippetDirectory`, but we removed it because it was very confusing to understand how it interplayed with the search path.

I am not super happy with your proposed solution either, because it seems magic that the first full path in the search directories should become the only displayed/used writing place for the snippets.

I feel bringing back a special configuration variable that if set defines which directory is used for UltiSnipsEdit is the best option, however I am unsure about its naming to not be confusing. Do you have a suggestion?

zhoosch:
Yes, i believe, a separate variable is a much more better solution than mine.

Maybe the names `g:UltiSnipsSaveDirectory` or `g:UltiSnipsWorkingDirectory` can do it?

SirVer:
How about `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit`? It is long, but seems quite distinct. If that variable is unset (default), we keep the current behaviour, also when `UltiSnipsEdit!` is used. If it is set and the command is used without a bang we default to using the first existing file that matches in that directory or create a new file `<ft>.snippets`. In neither case is the search path for snippet expansion changed.

What do you think?

SirVer:
@zhoosch Do you want to tackle these changes in this PR or should we close this and put this change into the backlog?

zhoosch:
Hi @SirVer,

the plan is good. I can tackle it. But i have a lot of stuff so i can do it in 2 weeks. If this time frame is not an issue i would change this PR here to fit the expected behavior.

A question: Is Backlog equal to Issues?

SirVer:
@zhoosch 2 weeks are no issue at all, this is not urgent.

Yes, Backlog is the same as issues. A Jiraism that sneaks in every once in a while from my site, sorry for the confusion.

PR #1309, opened 2021-01-29
Enhancing the wording of an example

kunzaatko:
1) Adding 'the' articles, where they are missing
2) Changing `<BS>` to <kbd>BackSpace</kbd> for consistency
3) Splitting into multiple sentences for easier reading


SirVer:
@kunzaatko Thank you very much!

PR #1314, opened 2021-02-05
Added CanExpandSnippet, CanJumpForwards, CanJumpBackwards

### LINKED ISSUE #786: Add UltiSnips#isExpandable()
Add a function UltiSnips#isExpandable() (as neosnippet offers in guise of neosnippet#expandable()), useful for example for autocompletion plug-ins like delimitmate:
```vim
 nnoremap <expr> <tab> delimitMate#isEmptyPair() && UltiSnips#isExpandable() ? <del>:UltiSnips#Expand() : '<tab>'
```
This way, a snippet with delimiters such as
```vim
  snippet [
	[${1:text}](http://${2:address})
```
in markdown.snippets of vim-snippets, is expanded with no trailing ] (auto-inserted by the autocompletion plugin).

thorlucas:
Closes #786.

Adds three functions, `UltiSnips#CanExpandSnippet`, `UltiSnips#CanJumpForwards`, and `UltiSnips#CanJumpBackwards`, which *significantly* cleans up more complex expansion functions.

The issue with the current method is that it requires we actually *run* the `<C-R>=UltiSnips#ExpandSnippet()<CR>` then check the result. This makes actually doing anything after it very difficult if you're in an `<expr>` mapping. Now we can have `<expr>` mappings that look like this: 

```lua
function expand_or_jump()
    -- ...
    if vim.api.nvim_eval([[ UltiSnips#CanExpandSnippet() ]]) == 1 then
        return [[ <C-R>=UltiSnips#ExpandSnippet()<CR> ]]
    end
    -- ...
end
```

and etc.

Issue #786 was previously closed because it was said that `UltiSnips#SnippetsInCurrentScope()` has the same functionality, but it actually doesn't as I pointed out in the comments.

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -94,6 +94,21 @@ function! UltiSnips#SnippetsInCurrentScope(...) abort
     return g:current_ulti_dict
 endfunction
 
+function! UltiSnips#CanExpandSnippet() abort
+	py3 vim.command("let can_expand = %d" % UltiSnips_Manager.can_expand())
+	return can_expand
+endfunction
+
+function! UltiSnips#CanJumpForwards() abort
+	py3 vim.command("let can_jump_forwards = %d" % UltiSnips_Manager.can_jump_forwards())
+	return can_jump_forwards
+endfunction
+
+function! UltiSnips#CanJumpBackwards() abort
+	py3 vim.command("let can_jump_backwards = %d" % UltiSnips_Manager.can_jump_backwards())
+	return can_jump_backwards
+endfunction
+
 function! UltiSnips#SaveLastVisualSelection() range abort
     py3 UltiSnips_Manager._save_last_visual_selection()
     return ""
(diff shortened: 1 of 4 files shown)


dbalatero:
@SirVer Would really love to see this as well, so I can use it in an `<expr>` mapping!

SirVer:
@thorlucas I reviewed, simplified the code a tiny bit and added documentation. I was thinking about how to test that, but it seems minor enough that I feel it is unlikely that this will break in the future. Thanks for your contribution!

thorlucas:
Awesome!

dbalatero:
Thanks @thorlucas and @SirVer! I have it integrated already with nvim-compe, snippets.nvim and UltiSnips:

https://github.com/dbalatero/dotfiles/blob/d87d2e6668e1d9dab9c935488c5fcdd78aa7ec84/nvim/lua/lspservers.lua#L93-L116

Issue #1252, opened 2020-06-23
Ultisnips cannot detect default user folder for NeoVim on Windows.


xarthurx:
**Expected behavior:**
I can edit my own snippets when using `:UltiSnipsEdit`.

**Actual behavior:**
A message pop up:

`UltiSnips was not able to find a default directory for snippets. Do you have a .vim directory? Try :UltiSnipsEdit! instead of :UltiSnipsEdit.`

**Steps to reproduce**

1. Install UltiSnips for Neovim using *vim-plug* on Windows 10.

2. type `:UltiSnipsEdit`.

PS.1: If I use  `:UltiSnipsEdit!`.
only the official snips from `vim-snippets` was listed.
![image](https://user-images.githubusercontent.com/1921878/85388366-51076a00-b546-11ea-9338-e7b399a709ce.png)

PS.2: I tried adding dirs and runtimes in the `init.vim` file, but things don't get better.
Is it because ultisnips only check the `$XDG_CONFIG` folder for linux machines, but not `$LOCALAPPDATA` where Nvim is installed on Windows?

-----

- **Operating System**:  Windows 10
- **Vim Version**:  Neovim latest.
- **UltiSnips Version**:  latest.
- **Python inside Vim**:  3.7
- **Docker repo/vimrc**:

xarthurx:
Additional notes:

If I set `g:UltiSnipsSnippetDirectories` to a dir with an absolute path, then I can edit my own snippets.
But once I include the keyword `UltiSnips` in that settings (or use it only as the default settings), the above issue will occur.

Kiniwara:
Same problem, vim-plug, win10

FruitDZign:
I have the exact same issue as xarthurx and Yonshoku (neovim v0.5.0-811 and then neovim v0.5.0-828 64 bits + Windows 10 20H2 + Python 3.7 + vim-plug)
Even when setting the absolute path of my snippets or adjusting the runtimepath variable I had the same error.

cristianUrbina:
You need to create a directory called UltiSnips inside ~/.vim directory
`mkdir ~/.vim/UltiSnips`

SirVer:
I have no idea what Neovim does differently so this does not work. Please try reproducing with Vanilla Vim. I'll leave the ticket open for a while until I see a repro with Vanilla Vim, otherwise I assume this is a neovim only problem.

[...]

PR #1328, opened 2021-04-02
Change the way var expansion is handled and change UltiSnipsEdit logic.

UmbralReaper:
#### Purpose of Change
fix https://github.com/SirVer/ultisnips/issues/1252
Properly handle variables like `%LOCALAPPDATA%` and `$HOME`. 

#### Describe the Solution
Remove the two lines that mandated that the snippets had to be in `.vim`.
Create a new general function, `expand_path` and use that wherever `os.path.expanduser` was previously used.

#### Additional Context
I would be happier if python 3.5 support could be dropped completely, and we use `pathlib.Path` for most of this instead. Is there a reason to keep 3.5 support around?


pythonx/UltiSnips/snippet_manager.py
```diff
@@ -882,8 +883,6 @@ def _file_to_edit(self, requested_ft, bang):
             # the users vim config directory.
             dot_vim_dir = vim_helper.get_dot_vim()
             for snippet_dir in all_snippet_directories:
-                if Path(dot_vim_dir) != Path(snippet_dir).parent:
```
SirVer, line 885: This was introduced in #1138. I think it was meant so that if there is a HOME dir for the user, it should be used for editing snippets preferably. You disable this, which makes sense for Windows, but not for most other OSes.

SirVer (commented):
Rest of the code looks good to me, only the removal of these two lines seems breaking to me.

SirVer:
I took the liberty to merging master into this PR to get the tests running again. 

I share your sentiment for Python 3.5. It is actually end of lived for a while now, but in #1208 it was pointed out that it was still the default Python in debian stretch which is supported till June 2022 and will hence still have quite some users.

SirVer:
If I run `:UltiSnipsEdit` with your changed branch, I see all suggestions from all plugins in my RTP: 

grafik

with `master` I only see the 2 actually existing files:

grafik

This change still has bugs therefore.

UmbralReaper:
Now it checks for a config dir (in an expanded list of locations) first, and only if that does not exist will it show the whole list.

SirVer:
@UmbralReaper the tests are failing. Is this ready for review?

UmbralReaper:
They shouldn't be failing. Just for now, I'll mark as draft.

SirVer:
@UmbralReaper I am going to close these for now, since you do not seem to be actively working on them. If you ever pick these up again, feel free to reopen!

PR #1347, opened 2021-04-17
Fix #1339 wrong cursor position in nested snippets

### LINKED ISSUE #1339: Mirrors cause wrong cursor position in nested snippets
**Expected behavior:**
The cursor should be advanced according to both the characters typed and the characters inside the mirrors of the current tabstop that precede it. 

**Actual behavior:**
The cursor is advanced additionally according to all mirrors, rather than just those preceding the text. This only happens inside nested snippets though and works fine at top-level. It is also noteworthy that this only seems to happen if the mirrors are not separated by any characters between them. If say a "b" is inserted between the two `$1` it works fine.

**Steps to reproduce**

1. Register the following snippets **OR** clone my reproduction fork linked below and run it with Docker
   ```
   snippet test "test"
   $1$1 $0
   endsnippet

   snippet outer "outer"
   $1 $0
   endsnippet
   ```
2. Open a new file with the snippets active and type `outer<tab>test<tab>a`
3. Observe that the cursor has moved behind both "a" characters that appeared, rather than having been placed between them.
-----

- **Operating System**: Manjaro 21.0 Ornara
- **Vim Version**: 
   NVIM v0.4.4
   Build type: Release
- **UltiSnips Version**: 3.2
- **Python inside Vim**: 
   3.9.2 (default, Feb 20 2021, 18:40:11)                                                                                                                                                                                              
   [GCC 10.2.0]
- **Docker repo/vimrc**: [reproduction fork](https://github.com/JeWe37/ultisnips)

JeWe37:
This PR fixes the problem mentioned in issue #1339. _VimCursor dummy NoneditableTextObject is now created as a child of the EditableTextObject actually being edited at that time. Thereby it is ensured that it is unambiguous which mirrors should and should not move the cursor position further. It also resolves general issues when a snippet contains only tabstops and mirrors and no actual text.

Tests have been added to ensure that this remains functional and the test suite completes successfully.

### DIFF
--- a/docker/snippets/all.snippets
+++ b/docker/snippets/all.snippets
@@ -1,7 +1,3 @@
-snippet test "test"
-$1$1 $0
-endsnippet
-
-snippet outer "outer"
-$1 $0
+snippet example "Repro case example"
+This is a simple snippet example
 endsnippet
--- a/docker/snippets/all.snippets
+++ b/docker/snippets/all.snippets
@@ -1,3 +1,7 @@
-snippet example "Repro case example"
-This is a simple snippet example
+snippet test "test"
+$1$1 $0
+endsnippet
+
+snippet outer "outer"
+$1 $0
 endsnippet
--- a/pythonx/UltiSnips/debug.py
+++ b/pythonx/UltiSnips/debug.py
@@ -21,13 +21,14 @@ with open(DUMP_FILENAME, "w"):
 def echo_to_hierarchy(text_object):
     """Outputs the given 'text_object' and its children hierarchically."""
     # pylint:disable=protected-access
+    orig = text_object
     parent = text_object
     while parent._parent:
         parent = parent._parent
 
     def _do_print(text_object, indent=""):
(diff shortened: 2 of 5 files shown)


SirVer:
Wow! I am very deeply impressed by you stepping up and figuring this out. This is very complex code and you seem to have understood it really well and really fast too. I do not think I could have fixed this bug as quickly as you did, for I have not touched this code in ages.

Thank you very much for your work and contribution! This is highly appreciated. You are not by chance looking for a tech job in Munich :) ?

The failing tests already time out on master. I have not yet investigated why.

JeWe37:
Hahaha, thanks, glad I could be of help, but I'm just some first year not-even-quite CS student in Aachen, so not currently looking for a job. And really it took me a bit to wrap my head around this code as well, but luckily the issue was fairly localized.

SirVer:
Thanks again! Good luck for your studies, and keep us in mind: Qwello also hosts interns!

Issue #1188, opened 2020-03-06
when use signature help of YCM , ultisnips can not jump next tabstops. YCM: set completeopt+=popup


shiqf:
filetype: python

config: 
    let g:UltiSnipsExpandTrigger       = "\<tab>"
    let g:UltiSnipsJumpForwardTrigger  = "\<c-j>"
    let g:UltiSnipsJumpBackwardTrigger = "\<c-k>"

snippet: 
snippet for "for loop" b
for ${1:item} in ${2:iterable}:
	${3:${VISUAL:pass}}
endsnippet

**Expected behavior:**

press \<c-j>enumerate()\<c-j> =>

for item in enumerate(): 
    pass|

**Actual behavior:**
 after type enumerate()  \<c-j> or \<c-k> cursor can not move

for item in enumerate()|: 
    pass

**Steps to reproduce**

1. for\<tab>.  =>
2. for item| in iterable: 
    pass
3. \<c-j>
4. for item in enumerate()|: 
    pass|
5. \<c-j>

![image](https://user-images.githubusercontent.com/41278084/76114178-7cf6f200-6020-11ea-93d5-1dda5626eb77.png)

-----

- **Operating System**:  
Mac OS 10.5 
- **Vim Version**: 
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Mar  4 2020 22:41:50)
macOS version
包含补丁: 1-350
- **UltiSnips Version**: 
e910b4fb9e276d18ed140fc492f30967ef9ace59
- **Python inside Vim**: 
3.7.6 (default, Dec 30 2019, 19:38:26)
[Clang 11.0.0 (clang-1100.0.33.16)]
- **Docker repo/vimrc**:

call plug#begin(get(g:, 'bundle_home', '~/.vim/bundles'))

    Plug 'ycm-core/YouCompleteMe', { 'do': 'python3 install.py --clangd-completer --ts-completer' }

        " 触发快捷键设置
        let g:ycm_key_list_select_completion   = ['<c-n>']
        let g:ycm_key_list_previous_completion = ['<c-p>']
        let g:ycm_key_list_stop_completion = ['<c-y>']
        let g:ycm_key_invoke_completion = '<c-z>'
        " 当用户的光标位于诊断行上时用于显示完整诊断文本。默认 <leader>d
        let g:ycm_key_detailed_diagnostics = '<leader>d'
        set completeopt+=popup

        " noremap <c-z> <NOP>

[...]

PR #1362, opened 2021-06-06
Fix for Issue #1188

bloop132435:
This pull request aims to fix issue #1188, by ignoring buffer changes in insert mode (ie the completion menu)

Screenshot from 2021-06-05 19-31-20

It doesn't seem to fail any tests, but there are two tests that are skipped.

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -154,6 +154,15 @@ function! UltiSnips#LeavingBuffer() abort
     let from_preview = getwinvar(winnr('#'), '&previewwindow')
     let to_preview = getwinvar(winnr(), '&previewwindow')
 
+    if "ic" == mode()
+        return
+    endif
+    if "i" == mode()
+        return
+    endif
+    if "ix" == mode()
+        return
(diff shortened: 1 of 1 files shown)


SirVer:
This will lead to other problems. The reason we terminate snippets when we leave a buffer is because UltiSnips internally has make a diff of buffers for each Cursor Movement to apply a heuristic to figure out where the tabstop positions moved. If the buffer changes, UltiSnips will get confused and apply a diff between the old and the new buffer. 

Since diffing is `O(n^3)` this can basically freeze up Vim if the original buffer is large and the preview buffer contains a decent amount of content. Another problematic case is using something like window movement commands that can be used in InsertMode to switch within buffers, the same thing would happen.

I think UltiSnips could be made smarter about ignoring certain types of buffers and not track cursor movement when inside these buffers. However this change needs to be done deeper in the logic of UltiSnips. But this proposed change cannot go in like this.

Thanks for contributing! I hope you can understand my reasoning for not merging this.

bloop132435:
That makes sense.

Issue #1393, opened 2021-09-21
[Feature request] Have a python persistant storage across snippet invocation (to memorize variable values)


hl037:
Though it is already possible using some hackery and files...
...It would be great to have a simple way to store values accessible accross snippet invocations.
There could be several level of storage : 
 - Snippet storage (persistent for the same snippet re-invokation)
 - Buffer storage (persistant for any snippet invokation in the same buffer)
 - Session storage (persistant for any snippet invokation in any buffer)
 - File storage (persistant for any snippet invokation in any vim instance for this file)
 - Global storage (persistant for any snippet invokation, in any vim instance).

Example of usecase (this is a personal snippet for Criterion test cases, the one the idea came to me) : 
```
snippet ft "test" b
TEST(${1:`!p snip.rv=snip.store.file["test_suite_name", ""]`}, $2){
	$0
}`!p
store.test_suite_name = t[1]
snip.rv=snip.store.file["test_suite_name"] = t[1]
`
endsnippet
```

...Then each time the snippet is called, the test suite's name will be the previous set.

This could also be usefull ofr C++ method implementation to fill the classname part to the previous one, and it would permit wonderful new logic and state-full templates...

hl037:
...Btw, I am able to implement this feature if you think it would be a good addition. But I would also like to discuss the interface. I wrote the example thinking "How would I expect it to be used", but I curious about other inputs =)

SirVer:
This is a complex feature, pretty hard to explain to new users. I see the appeal of it though, I have always build similar things using hackery in my `global` sections in the snippet files. I am not against adding this feature, providing it comes with good tests & documentation.

SirVer:
Closing as out of scope. Re-reading this years later, the proposal is effectively a five-tier scoped key-value store (`snippet` / `buffer` / `session` / `file` / `global`) — that's a meaningful new public API surface, with its own edge cases around `:saveas`, `:bwipeout`, `:mksession`, undo, etc., and it duplicates ground Vim already covers (`b:` / `t:` / `g:` plus `shada` / `viminfo`).

Everything in the listed scopes is achievable today with existing escape hatches:

- **Per-snippet** state: `snip.context` or a module-level dict in a `global !p … endglobal` block.
- **Per-buffer / per-tab / global session** state: write to `b:` / `t:` / `g:` from Python via `vim.eval` / `vim.command`.
[...]

PR #1394, opened 2021-09-22
Added Stores for python interpolations. Implements #1393

hl037:
(See the diff in the doc for a full explanation)

Adds dict-like stores to the python interpolation `snip` object to implement #1393 : 

- `snip.store.tmp` : a store reset at each snippet run, but that stay alive across a snippet updates. Accessible only on the same python code part
- `snip.store.snippet` : a store reset at each snippet run, but that stay alive across a snippet updates. Accessible in all python code parts
- `snip.store.buffer` : a store common to all snippet executing in a same buffer
- `snip.store.session` : a store common to all snippet in a same vim instance
- `snip.store.file` : Same as buffer, but persisted to hard drive, and synced with other vim instances
- `snip.store.common` : Same as session, but persisted to hard drive, and synced with other vim instances

In addition a g:UltiSnipsStoreDir is added choose the directory where the persistent stores are saved. If this variable is not provided by the user, the stores are not persisted

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -25,7 +25,8 @@ UltiSnips                                      *snippet* *snippets* *UltiSnips*
       3.4.4 UltiSnips#CanExpandSnippet          |UltiSnips#CanExpandSnippet|
       3.4.5 UltiSnips#CanJumpForwards           |UltiSnips#CanJumpForwards|
       3.4.6 UltiSnips#CanJumpBackwards          |UltiSnips#CanJumpBackwards|
-   3.5 Missing python support                   |UltiSnips-python-warning|
+   3.5 Store persistence directory              |UltiSnips-store-directory|
+   3.6 Missing python support                   |UltiSnips-python-warning|
 4. Authoring snippets                           |UltiSnips-authoring-snippets|
    4.1 Basics                                   |UltiSnips-basics|
       4.1.1 How snippets are loaded             |UltiSnips-how-snippets-are-loaded|
@@ -492,13 +493,15 @@ This function returns 1 if UltiSnips can jump backwards in the current
 situation. This is useful in conditional mappings.
 
 
-3.5 Store persistence directory                         *g:UltiSnipsStoreDir*
+3.5 Store persistence directory                   *UltiSnips-store-directory*
 -------------------------------
+                                                        *g:UltiSnipsStoreDir*
+Some interpolation data stores need persistence ('file' and 'common' stores).
+The g:UltiSnipsStoreDir should to be set to an absolute path string pointing
+to an existing directory to provide this functionality. If not set, or set
+to v:null, these stores will behave like 'buffer' and 'session'. >
 
-Some interpolation data stores need persistence ('file' and 'common' sotres).
-The g:UltiSnipsStoreDir should to be set to an existing directory to provide
-this functionnality. if not set, or set to v:null, these stores will behave
-like 'buffer' and 'session'.
+   let g:UltiSnipsSnippetDirectories="~/.vim/UltiSnips/stores"
 
 
 3.6 Warning about missing python support           *UltiSnips-python-warning*
@@ -509,9 +512,9 @@ python support.  If no support is detected, a warning will be displayed and
 loading of UltiSnips will be skipped.
 
 If you would like to suppress this warning message, you may add the following
-line to your vimrc file.
+line to your vimrc file. >
 
-    let g:UltiSnipsNoPythonWarning = 1
+   let g:UltiSnipsNoPythonWarning = 1
 
 This may be useful if your Vim configuration files are shared across several
 systems where some of them may not have Vim compiled with python support.
@@ -970,12 +973,13 @@ snippet executions: >
 
     snip.store.tmp:
         A temporary local store accessible only during the snippet execution
-        and updates in the same python code section. It is reset each time the snippet is called, but not
-        between updates.
+        and updates in the same python code section. It is reset each time the
+        snippet is called, but not between updates.
 
     snip.store.snippet:
         A temporary local store accessible only during the snippet execution
-        and updates in all python code sections. It is reset each time the snippet is called, but not
+        and updates in all python code sections. It is reset each time the snippet
+        is called, but not
         between updates.
 
     snip.store.buffer:
@@ -992,13 +996,21 @@ snippet executions: >
         similary to store.buffer.
 
     snip.store.common:
-        A persistent global store common to all snippet executions in all buffers and any
-        vim instance. If g:UltiSnipStoreDir is unset or set to v:null, then it behave
-        similary to store.session.
+        A persistent global store common to all snippet executions in all buffers
+        and any vim instance. If g:UltiSnipStoreDir is unset or set to v:null,
+        then it behave similary to store.session.
 
-The stores have some syntax sugars compared to built-in dicts : >
+Note : The stores are persisted as json. Keys and data stored are subject to the
+same limitations as the `json` python module (keys should be strings, and data
+should be made of 'None', 'bool' 'int', 'float', 'string', 'list' and 'dict').
+
+Note : Since JSON is a plain text file format, snippets should not store any
+sensitive information. If you are using this feature, please advise your users
+of this risk (e.g. Database passwords, private keys etc.).
+
+The stores have some syntax sugars in addition to built-in dict interface : >
     store['key', default]
-      where default is either a value either a factory is equivalent to:
+      where default is either a value either a factory, is equivalent to:
       if key not in store:
         store['key'] = default
       return store['key']
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -491,7 +491,17 @@ situation. This is useful in conditional mappings.
 This function returns 1 if UltiSnips can jump backwards in the current
 situation. This is useful in conditional mappings.
 
-3.5 Warning about missing python support           *UltiSnips-python-warning*
+
+3.5 Store persistence directory                         *g:UltiSnipsStoreDir*
+-------------------------------
+
+Some interpolation data stores need persistence ('file' and 'common' sotres).
+The g:UltiSnipsStoreDir should to be set to an existing directory to provide
+this functionnality. if not set, or set to v:null, these stores will behave
+like 'buffer' and 'session'.
+
+
+3.6 Warning about missing python support           *UltiSnips-python-warning*
 ----------------------------------------
 
 When UltiSnips is loaded, it will check that the running Vim was compiled with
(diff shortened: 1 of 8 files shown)


hl037:
...I do understand I may need to add some tests before it is validated... But at least, you can try out the feature =)

hl037:
Another feature I would like to add is a shortcut to define memorized tab values...

In my snippet definitions, I use this global python code : 

```
_once = None

def once():
	global _once
	if _once is None:
		_once = T['once', True]
		T.once = False
	return _once

class StoreReadProxy(object):
	def __init__(self, tab, store):
		self.store = store
		self.tab = tab
	
	def __getitem__(self, key):
		if not isinstance(key, tuple):
			key = key, ''
		rv = self.store[key]
		t[self.tab] = rv
	
	def __getattr__(self, key):
		return self.__getitem__(key)

class StoreWriteProxy(object):
	def __init__(self, tab, store):
		self.store = store
		self.tab = tab
		
	def __getitem__(self, key):
		if isinstance(key, tuple):
			key, _ = key
		self.store[key] = t[self.tab]
		
	def __getattr__(self, key):
		return self.__getitem__(key)

def mem_tab(tab, store):
	if once():
		return StoreReadProxy(tab, store)
	else :
		return StoreWriteProxy(tab, store)
```

...It allow to easily implement persistent tab default value... example : 

```
snippet ptd "Parameterized test data" b
DECLARE_PARAMS_SECTION(${1:`!p mem_tab(1, snip.store.buffer).params_section`}){

DECLARE_PARAM_TYPE{
	const char * name;
	$2
};

DATA_CASE_TEST(PARAM_TYPE & param, $3){
	$0
}

};

endsnippet

snippet pt "Parameterized test" b
BEGIN_PARAMS_SECTION(${1:`!p mem_tab(1, snip.store.buffer).params_section `})
	DATA_CASE($2)$0
END
endsnippet
```

...That permits to propose a default tab value according the the previous call to `pt` or `ptd`

doc/UltiSnips.txt
```diff
@@ -491,7 +491,17 @@ situation. This is useful in conditional mappings.
 This function returns 1 if UltiSnips can jump backwards in the current
 situation. This is useful in conditional mappings.
 
-3.5 Warning about missing python support           *UltiSnips-python-warning*
+
+3.5 Store persistence directory                         *g:UltiSnipsStoreDir*
+-------------------------------
+
+Some interpolation data stores need persistence ('file' and 'common' sotres).
```
SirVer, line 498: typo: sotres

doc/UltiSnips.txt
```diff
@@ -491,7 +491,17 @@ situation. This is useful in conditional mappings.
...
+The g:UltiSnipsStoreDir should to be set to an existing directory to provide
```
SirVer, line 499: please be precise here about type of variable (probably string?), also the path should be absolute.

doc/UltiSnips.txt
```diff
@@ -491,7 +491,17 @@ situation. This is useful in conditional mappings.
...
+this functionnality. if not set, or set to v:null, these stores will behave
```
SirVer, line 500: type: functionality

doc/UltiSnips.txt
```diff
@@ -491,7 +491,17 @@ situation. This is useful in conditional mappings.
...
+like 'buffer' and 'session'.
+
```
SirVer, line 502: add a paragraph, that snippet persistent information is saved in plain text, i.e. no sensitive information should be contained.

doc/UltiSnips.txt
```diff
@@ -491,7 +491,17 @@ situation. This is useful in conditional mappings.
 This function returns 1 if UltiSnips can jump backwards in the current
```
SirVer, line 491: probably need to update the table of contents at the beginning of the file.

doc/UltiSnips.txt
```diff
@@ -954,6 +964,49 @@ The 'snip' object provides some properties as well: >
...
+        store. If g:UltiSnipStoreDir is unset or set to v:null, then it behave
+        similary to store.buffer.
+
+    snip.store.common:
+        A persistent global store common to all snippet executions in all buffers and any
+        vim instance. If g:UltiSnipStoreDir is unset or set to v:null, then it behave
+        similary to store.session.
+
+The stores have some syntax sugars compared to built-in dicts : >
+    store['key', default]
```
SirVer, line 1000: I find the syntax for default unusual and unexpected. I'd suggest to instead use `get` to keep symmetry with Python's dictionary. The downside is, that the access to the store now needs 4 strokes more. Maybe we can come up with a shorter form instead of `snip.store.<type>`

doc/UltiSnips.txt
```diff
@@ -981,6 +1034,37 @@ endsnippet
...
+sure the counter is incremented only once (else, the snippet update would run the
+code again)
+
+------------------- SNIP -------------------
+snippet inc "increment"
+`!p
+if snip.store.tmp['todo', True]:
+    snip.rv = snip.store.file['i', 1]
+    snip.store.file.i += 1
+    snip.store.tmp.todo = False
```
SirVer, line 1060: I disagree with this shortcut, let's keep the thing a dictionary and not get fancy with accessors.

pythonx/UltiSnips/snippet_manager.py
```diff
@@ -144,6 +145,8 @@ def __init__(self, expand_trigger, forward_trigger, backward_trigger):
         self._should_update_textobjects = False
         self._should_reset_visual = False
 
+        self._storeManager = StoreManager()
```
SirVer, line 148: consistency: use `_store_manager` as everywhere.

pythonx/UltiSnips/store.py
```diff
@@ -0,0 +1,237 @@
+from UltiSnips import vim_helper
+from pathlib import Path
+import json
+from collections import defaultdict
+
+class Store(object):
+    """
+    A dict-like supporting supporting the obj[key, default] synthax
```
SirVer, line 8: typo: synthax

pythonx/UltiSnips/store.py
```diff
@@ -0,0 +1,237 @@
...
+    @classmethod
+    def get(cls):
+        if cls._store is None :
+            cls._store = cls()
+        return cls._store
+
+
+class _PersistentStore(Store):
+    """
+    A Store persistent saved to json format.
```
SirVer, line 114: this is important information: you can only story stuff that is serializable to JSON. You should notice this in the docs or use 'pickle' to have fewer (but still some) limitations.

pythonx/UltiSnips/store.py
```diff
@@ -0,0 +1,237 @@
...
+            with open(self._path, 'w') as f :
+                json.dump(self._dict, f)
+
+    def __setitem__(self, key, val):
+        self._changed = True
+        super().__setitem__(key, val)
+
+
+def _encodeFilePath(path):
+    return str(Path(path).resolve()).replace('%', '_%%_').replace('/', '%')
```
SirVer, line 139: what are the `%` doing? This seems to be unix focused, will this work on Windows as expected?

pythonx/UltiSnips/store.py
```diff
@@ -0,0 +1,237 @@
...
+
+def _storeDir():
```
SirVer, line 141: style: naming not camel case

SirVer (commented):
Code is pretty neat, well done. I had a few thoughts and comments and it also still needs some tests.

pythonx/UltiSnips/snippet_manager.py
hl037: Oh yeah sorry, I corrected the indent but forgot about the case I will add some commits anyway

pythonx/UltiSnips/store.py
hl037: I agree, I used json because pickle would leave a visious security breach, because one would just have to put malicious code in the common store file to make it executed by vim (and since vim could be used as root by some users...)

I have to better document it

pythonx/UltiSnips/store.py
hl037: The `.replace('/', '%')` encodes the "/" characters to avoid creating subdirs (in a similar way vim does for the swap files). Since it's not a reserved name character in neither posix nor windows, it should work (except may be the file name length limit, that may need specific testing with FAT32... I may look at how vim handles it with swap files).

the `.replace('%', '_%%_')` simply escapes the '%' character to make sure any file (even containing '%') has a unique file path. The pattern is complex to handle cases like "/path%/tofile". I don't really know why one would put "%" in a file or dir name... but you know, "always consider the user dumber than the computer".

doc/UltiSnips.txt
hl037: Sorry for the typos, it was late when I wrote the doc, I actually just tried to convey what I had in mind the best possible and I admit I haven't reread carefuly enough

doc/UltiSnips.txt
hl037: Good observation !

hl037:
Thank you very much for the comments and the compliment ! I am glad you like the feature.
I will rework it, and read a bit more how the tests are performed (...that scared me a bit actually so that I delayed this topic until being sure I have green light :D )

...Well, when I say "scare", it's more about the docker and tmux automation that looks "Wow" but I actually think it's really neat to do the tests like this and I was surprised how tests are actually readable in the end

doc/UltiSnips.txt
SirVer: no worries, I would not have polished at the draft state either.

pythonx/UltiSnips/store.py
SirVer: That is fragile, at windows you would need to replace '\\' as path separator and maybe other characters. How about you just sha1 hash the resolved path, you will not have any problems with special characters and a flat hierarchy with no collisions like this, however you cannot direct back from hash to filename. Which might be considered a slight security plus.

Alternatively you could use base32 to encode bijectively between paths and save filenamen (see https://stackoverflow.com/questions/3945541/base64-encoding-safe-for-filenames).

SirVer:
The tests are not as scary once you look into them, they are just typing out what you as a manual test would type out too. Docker is not necessary, you can just as well run a tmux session on your machine, but I find running in docker easier. 

I will be very busy the next few days, so I might not be as fast to review as I'd like to be. But please ping me once you have done another pass and deem it ready.

SirVer:
@hl037 I did not have another look at the code yet, assuming that you are still working on it. Let me know if that is incorrect.

hl037:
I have to add a `._after_test` method in the test case to check the creation of the json file. While I looked at the code, I though of a pretty neat way to add supporting multi run while warrenting previous test will behave exactly the same :

This code ( vim_test_case.py:189 ) would be put in its own function taking the config and the key strokes
```
        self.in_vim_python_version = self.vim.launch(vim_config)

        self._before_test()

        if not self.interrupt:
            # Go into insert mode and type the keys but leave Vim some time to
            # react.
            text = "i" + self.keys
            while text:
                to_send = None
                for seq in SEQUENCES:
                    if text.startswith(seq):
                        to_send = seq
                        break
                to_send = to_send or text[0]
                self.vim.send_to_vim(to_send)
                time.sleep(self.sleeptime)
                text = text[len(to_send) :]
            self.output = self.vim.get_buffer_data()
        self._after_test()

```

...To have a multi run test, first, set keys and wanted as a tuple of strings (one value per run)
Then, 2 solutions : 
1) creating a new class that implement the "do_test" (or other name) function (the one created from the refactoring of the previous code block) by calling super() for each run, and setting output to a tuple too. This way, the code checking the output does not need change since tuple equality comparison will work deeply in the tuple (and single run tests will still be comparing string in the current way).

2) The same, but single / multi-run is detected by checking if the keys and wanted fields are str or tuple.

I prefer the 2) because it provides the same interface for all cases, and the low amont of modification  (and I could also formally prove the execution path for the single run case is equivalent to the current implementation).

Anyway, I won't touch this in the current PR. I think I will only write simple tests only for the `common` store (since the `file` work along the same principle, but need more work with the hash since it's not trivial to import the hash function from the test script

hl037:
Remains only the tests for the reloading the `common` store from file, and saving.
Then also a last typo double check and it will be good for me...

hl037:
Well, I seek your help on a very weird "bug" when using the test suite with tmux...
first, try clone this branch : https://github.com/hl037/ultisnips/tree/store_

run `python ./test_all.py -v -i Store_SnippetSimple` So that the tmux session shows a vim instance with a "test" snippet declared.

There is a very strange thing happening : sometimes, when you trigger the snippet (`test\t`), the cursor ends up at the end of the expansion, sometimes at the begin.

It seems like if you press "o" to enter insert mode before invoking the snip, the cursor goes at the end as it should, but if you leave insert mode, go back with `I` to insert at the line begin, erase with <kbd>del</kbd> and perform again the snip, then the cursor goes at the begin of the line... This is quite random, it's the way I can reproduce almost each time, but you may have to try many time to get it to show the strange behavior...

I have absolutely no idea of what could cause that...

SirVer:
I think your idea about reusing the existing test infra and accepting a tuple as keys is fine, but let's discuss that in detail should you chose to implement this.

Let me know when I should have a deeper look again.

SirVer:
@hl037 Leo, I am going to close your pull requests now. While I am still very interested in having them, they have not seen any progress for a long time and I think it is time to close them. Once you pick up working on them again, I will gladly reopen and have a look!

Issue #1403, opened 2021-11-02
Python code order may be non-deterministic


hl037:
Still struggling with #1394 's tests and #1402, I think I have found yet another problem : 
Python interpolation seems not to be fully deterministic (despite the sort applied to the code objects before executing them).

with this template : 
```
snippet test4 "test"
`!p import sys;sys.modules['dummy'] = 'tt';snip.rv=''``!p snip.rv=sys.modules.get('dummy', '')`
endsnippet
```

I managed to get :

![image](https://user-images.githubusercontent.com/2771764/139830653-b4498b33-f92c-4421-b3f8-06732c35ad6e.png)

...It's really hard to reproduce, I had to keep expanding templates for hours to trigger the bug only sometimes, and just adding a line or two to the python code makes it not to be triggered...

I think using sets to tag the code objects as done / undone is quite obscure, and I would not be surprised if some of my struggles would come from `SnippetInstance.update_textobjects`

hl037:
...After successfully added a remote debugger (I think these 50 lines with no opther dependencies deserves also a pull request btw...) I pinpointed the problem :

![image](https://user-images.githubusercontent.com/2771764/139841357-00d21e4c-a3b5-4c00-9484-070083b2e0ba.png)

Clearly, there is no way to set an order on these two code objects, since the position and the tiebreaker are strictly the sames...

SirVer:
This is a great find. I think the debugger + some docu on how to use it would indeed be a very valuable addition, issues n UltiSnips are notoriously hard to debug some times. I think a new tiebreaker (or maybe even a better tiebreaker) could be the original start of a text object in the original snippet text. This can be provide by the lexer. After all, there is really no affordance to change the order of text objects within a snippet ever, and every text object has a length in the original snippet text.

hl037:
I don't know if it's a good idea to do major changes to the tie breaker, it may become a code breaker...
Before I made the PR for the debugger, I tried to simply add a z component to the Position class to order 0-length text object.
However, while reimplementing the move method : 

```
    def move(self, pivot, delta):
        """'pivot' is the position of the first changed character, 'delta' is
        how text after it moved."""
        if self < pivot:
            return
        if delta.line == 0:
            if self.line == pivot.line:
                self.col += delta.col
        elif delta.line > 0:
            if self.line == pivot.line:
                self.col += delta.col - pivot.col
            self.line += delta.line
        else:
            self.line += delta.line
            if self.line == pivot.line:
[...]

PR #1404, opened 2021-11-02
Added remote debug facilities (remote pdb server)

hl037:
As discussed in #1403, and since you seemed enthusiast about it, I polished the remote debugger / post-mortem in something quite usable even though some stuff are still missing (like handling correctly arrow and completion keys...)

I am a bit sorry the code is somewhat messy, it was not that easy to make sure issuing a quit command would close the socket with vim resuming as if no debugger were launched, but at least, I tested it with pdb and pdb++ with all configuration possibility and it does what it is asked for, without altering UltiSnips when not enabled (I ran the test suite, and I have no fails).

Please don't ask me to provide a test case, I have truly no idea of how I would implement it =D

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1888,13 +1888,15 @@ You can install it this way: >
 
 It is a no-configuration replacement of the built-in pdb.
 
-To connect to the pdb server, simply run telnet: >
+To connect to the pdb server, simply use a telnet-like client.
+To have readline support (arrow keys working and history), you can use socat: >
 
-    telnet host:port
+    socat READLINE,history=$HOME/.ultisnips-dbg-history TCP:localhost:8080
 
+(Change `localhost` and `8080` to match your configuration)
 To leave the server and continue the execution, run Pdb's `quit` command
 
-Known issues : The arrow and tab keys (for completion) are not supported yet
+Known issue: Tab completion is not supported yet.
 
 =============================================================================
 7. FAQ                                                        *UltiSnips-FAQ*
@@ -1921,7 +1923,7 @@ A: Yes there is, try
   endfunction
 
 =============================================================================
-7. Helping Out                                            *UltiSnips-helping*
+8. Helping Out                                            *UltiSnips-helping*
 
 UltiSnips needs the help of the Vim community to keep improving. Please
 consider joining this effort by providing new features or bug reports.
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -56,8 +56,6 @@ UltiSnips                                      *snippet* *snippets* *UltiSnips*
    5.1 Existing Integrations                    |UltiSnips-integrations|
    5.2 Extending UltiSnips                      |UltiSnips-extending|
 6. Debugging                                    |UltiSnips-Debugging|
-   6.1 Setting breakpoints                      |UltiSnips-breakpoints|
-   6.2 Accessing Pdb                            |UltiSnips-Pdb|
 7. FAQ                                          |UltiSnips-FAQ|
 8. Helping Out                                  |UltiSnips-helping|
 9. Contributors                                 |UltiSnips-contributors|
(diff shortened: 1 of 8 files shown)


SirVer:
Thanks! I'll try to review this weekend, I am quite busy though these days, might take me a bit longer.

hl037:
I found a way to have at least an history and the arrow keys (no completion though) :
using `socat READLINE,history=$HOME/.ultisnips-dbg-history TCP:localhost:8080`

doc/UltiSnips.txt
```diff
@@ -55,9 +55,12 @@ UltiSnips                                      *snippet* *snippets* *UltiSnips*
 5. UltiSnips and Other Plugins                  |UltiSnips-other-plugins|
    5.1 Existing Integrations                    |UltiSnips-integrations|
    5.2 Extending UltiSnips                      |UltiSnips-extending|
-6. FAQ                                          |UltiSnips-FAQ|
-7. Helping Out                                  |UltiSnips-helping|
-8. Contributors                                 |UltiSnips-contributors|
+6. Debugging                                    |UltiSnips-Debugging|
```
SirVer, line 58: This is a great writeup, but I feel it is out of scope for the end user documentation. Let's move this into a separate document or the contributing document.

pythonx/UltiSnips/err_to_scratch_buffer.py
```diff
@@ -4,9 +4,12 @@
 import traceback
 import re
 import sys
+from bdb import BdbQuit
 
 from UltiSnips import vim_helper
 from UltiSnips.error import PebkacError
+from UltiSnips.remote_pdb import RemotePDB
+import time
```
SirVer, line 12: nit: move into first import block (std library)

pythonx/UltiSnips/remote_pdb.py
```diff
@@ -0,0 +1,109 @@
+import sys
+from bdb import BdbQuit
+
+from UltiSnips import vim_helper
+import threading
```
SirVer, line 5: nit: move to first import block

pythonx/UltiSnips/remote_pdb.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _shutdown(self):
+        if self._pdb is not None :
+            import socket
+            self.connection.shutdown(socket.SHUT_RDWR)
+            self.connection.close()
+            self.server.close()
+            self._pdb = None
+
+    @classmethod
```
SirVer, line 65: Some of these @classmethods are not using `cls`, hence they should be stand alone methods in this module for minimized coupling.

pythonx/UltiSnips/remote_pdb.py
```diff
@@ -0,0 +1,109 @@
...
+    def is_enable(cls):
+        return bool(int(vim_helper.eval('g:UltiSnipsDebugServerEnable')))
+
+    @classmethod
+    def is_blocking(cls):
+        return bool(int(vim_helper.eval('g:UltiSnipsPMDebugBlocking')))
+      
+
```
SirVer, line 81: nit: two emtpy lines? Run black over the code?

pythonx/UltiSnips/remote_pdb.py
```diff
@@ -0,0 +1,109 @@
...
+    @classmethod
+    def breakpoint(cls, host=None, port=None):
+        if cls.singleton is None and not cls.is_enable() :
+            return
+        cls._create()
+        cls.singleton.start_server()
+        cls.singleton.set_trace(sys._getframe().f_back)
+
+    @classmethod
+    def pm(cls):
```
SirVer, line 96: what does `pm` stand for? Can this method get a more descriptive name?

SirVer (commented):
Only a few small comments, great work!

doc/UltiSnips.txt
hl037: I had the same though as you first, but then, I though that it could actually be useful when authoring snippet. Also, since there are user configurable options, having them into the doc permits to tag them (suppose you use it in your .vimrc to permanently set the host and port, and later you ask yourself what these variables are. With it in the doc, you can simple type `:help g:UltiSnips***`). If you still think it should not be in the user doc, then I can either move it to another help file, either put it in contributing

pythonx/UltiSnips/remote_pdb.py
hl037: It is to permit subclassing, and also to to group all the RemotePdb related method to the same class. This way, when other methods call them, we know immediately we just have to look in the same class. Though, they could be made staticmethod

pythonx/UltiSnips/remote_pdb.py
hl037: It's the standard name from the std lib for "Post Mortem" : https://docs.python.org/3/library/pdb.html#pdb.pm
I made this class such both the most common method from Pdb are available with the same name. I could add a line of documentation though

doc/UltiSnips.txt
SirVer: I think it is a far shot to expect a user of UltiSnips to set up debugging through pdb. I consider myself a Python expert and only very very rarely used pdb before, so I think this is not a strong argument.

As for the tagging: That is understandable, but then again the variable names are descriptive enough that I would guess that they are not for the general public to use. I would search through the UltiSnips repo to figure out what they are. I hence still feel that this does not belong in this document, please move.

pythonx/UltiSnips/remote_pdb.py
SirVer: +1 for staticmethods for sure, however I still think they should be standalone: I respect the argument that they are grouped through the class. But then again, they are also grouped through the module name which is fairly descriptive. Making them stand alone strongly communicates that they do not need to know anything about the class or its internal setup, also reducing the API surface of the class making it more granular and hence easier to change in the future.

I do not feel super strongly, I can live with staticmethods, but I'd prefer them to be standalone.

pythonx/UltiSnips/remote_pdb.py
SirVer: +1 for documentation, but with your explanation the name is correct here. I think the standard library made the mistake to name the function as such, but it is correct to follow the standard library here instead of improving on the naming I think.

doc/UltiSnips.txt
hl037: That was an unexpected response since I use pdb in a daily base, and it was the first thing searched for in the documentation after I learned about python interpolation (without success hence this PR).

Is it ok if I create a new vim doc file named Ultisnips-advanced.txt ?

SirVer: > That was an unexpected response since I use pdb in a daily base, and it was the first thing I searched for in the documentation after I learned about python interpolation

One thing I learned in my 30+ years in computer science is that different people gravitate towards different tools. I have always been less productive in IDEs, but had collegues that could only work with them. I never had a lot of use for debuggers, thinking and reading through code usually helped me the most - you are clearly different there. I am therefore mostly enthusiastic about this change, because it is orthogonal to my workflow and hence might make it easier for people that do not work like me - this will make UltiSnips more accessible for other devs.

> Is it ok if I create a new vim doc file named Ultisnips-advanced.txt, and adding a small "debugging" paragraph pointing to this file ?

Yes, sure. I think it would be good to still reference it in the contributing doc though.

SirVer:
@hl037 Léo, some commits seem to have gone missing. Could you please do not rebase within a pull request? In the end the whole PR will be squashed to a single commit anyways, and having the full history makes it much easier to see the evolution of this PR towards its eventual final landing.

And again, my apologizes for being so slow to respond lately. Live is busy lately.

hl037:
> @hl037 Léo, some commits seem to have gone missing. Could you please do not rebase within a pull request? In the end the whole PR will be squashed to a single commit anyways, and having the full history makes it much easier to see the evolution of this PR towards its eventual final landing.
> 
> And again, my apologizes for being so slow to respond lately. Live is busy lately.

Normally, no commit disappeared, I always keep a backup of all my branches before rebasing (because yes, it's pretty easy to mess up with it). The rebase was to keep the history logical, it's the way I work generally, trying to publish public branch with a clean history permitting to read what have been done, while keeping the original history not too far if a problem arise. If you prefer I push small fixup commits, I'm ok too

doc/UltiSnips.txt
hl037: Ok, I will change that tonight. It is already referenced in the contributing doc.

I like this debate. I agree with you, thinking an reading the code is the best. My use for the debugger is mainly to 1) stop right at the error and quickly inspect the variables to see if there is something wrong, and then read the code to know why, 2) When I start working on an existing code base, it's easier to understand the execution flow with breakpoints. I think the best is to try as much workflow as we can in our life so that we can select what works best for us and also understand how others thinks and work to ease communication :)

SirVer: Amen!

SirVer:
> The rebase was to keep the history logical, it's the way I work generally, trying to publish public branch with a clean history permitting to read what have been done, while keeping the original history not too far if a problem arise.

Again, this is probably stylistic. From my experience working in very different environments I came to the conclusion that the main task of the merge request history is to allow people understand the original proposal and its evolution over the course of the discussion. This is best done by never rebasing and allowing the history to show the real evolution, no matter how messy it becomes. Rebasing breaks tools and reviewers, for example some of the above links of our prior comments are no longer working since the SHAs they have been connected too have vanished.

Once the merge request goes to master, it will be a single commit anyways, so the history is not important for the evolution of the main line. 

So I do not mind if you do large or small fixups, just work in any granularity you like - but please ideally just add commits to the evolution of the merge request.

SirVer:
On another topic, do you have an idea why some tests fail? It seems unrelated to your changes, but also quite strange.

hl037:
Tests fail because I used `f''` formatted string available since python 3.6. On 3.5, the raises a synthax error... But I don't know why the test suite freeze.

If needed, I could use .format instead, but I really do think dropping python < 3.6 support would be the best...

At least, ubuntu 18.04 already use 3.6.x, and ubuntu 16.04 is getting REALLY old (at least, I guess only servers and old administrations computers would be concerned, and I would not recommend running vim with plugins on the former, and I guess users of the latter wouldn't bother using vim at all...)

SirVer:
> If needed, I could use .format instead, but I really do think dropping python < 3.6 support would be the best...

This should be a separate discussion and a separate MR, let's not conflate this with this feature.

SirVer:
Not too concerned about the freezing, we seem to have this on master already. Of course it would be better if the test suite actually was green, but that is definitely not your problem :)

Let's get the 3.5 compatibility reestablished and do a removal of 3.5 in another merge request if you feel strongly about it.  I am willing to see if people still care.

SirVer (approved):

hl037:
Done.
Since this message would only be 5 character long, a small teaser : I've found a way to report the exactely where in the snippet an error occured, and make pdb show the snippet python code block... But it will be for another thread/PR

SirVer:
Thanks for this contribution and sorry for my slowness!

Issue #1407, opened 2021-11-06
Tab navigation discussion


hl037:
To fix #1405 , I need to make some change some about the tabstops.

However, the current tab navigation is... well to be honest quite messed up...
...If you nest a certain amount of tab it even doesn't work : 
```
class TabStop_Attack(_VimTest):
    snippets = ("test", "${1:foo${4:${2:zzz}bar$3fo}}")
    keys = ""
    wanted = ""
```

I started to work on that too, but I see no reference for how it should work (except the test suite, with come quite convoluted example ones !)

...I propose this : 

   1) As long as the user makes no changes to the document, no matter the cursor movements, all the snippets are kept alive.

   2) If the user makes a change in a tabstop, all snippets that are not child of the tabstop remain a live

   3) When the user make a change outside of a tabstop, this happens : 
       3.1) The text objects which their span contains the change are all removed (from change tracking only) up to (but excluding) the first `SnippetInstance` ancestor.
       3.2) All SnippetInstance descendent of the first SnippetInstance are removed (from change tracking only) too

   4) If there the change is outside the root SnipetInstance, Ultisnips Tears down (All snippets are ended)

   5) Tabstop navigation follows a DFS, parent first then children, order in the alive snippets on the tabstop number, with tabstop 0 last.

First, even if this algorithm is not adopted, adopting a reference algorithm seems key to me to write pertinent test cases.
Then this algorithm in particular has several benefits : 

1) for non recursive snippet expansion, it folllows exactly the current behavior (this means that 80% of snippet working today are unaffected).

2) It keeps alive the snippets longer, allowing one to scroll and go back without the snippet killed (as long as no changes are made elsewhere)

3) The code to implement it is much simpler than the one currently used (I already made some work that do that except the snippet cancellation)

4) With the upcoming store feature, it allows much more powerfull logic, for example, automatically adding function arguments based on nested template etc.

5) Less bug-prone. This algorithm is described in 7 points, that is 10 lines, whici is in my (subjective I admit) opinion much simple to than to describe what is currently done

SirVer:
I fully agree with your reasoning. I do think your current outlined algo is close enough to intuition and to what is happening currently that adopting it would not disrupt too many users use case. Go for it, I'd say!

SirVer:
This bug has been fixed on master, likely by interplay between #1613 and #1623.

Issue #1405, opened 2021-11-04
Untrackable bug with cursor end position


hl037:
I though I was done with #1403 by using https://github.com/hl037/ultisnips/tree/tiebreaker

...But there is yet another "wtf" problem with the cursor position. And That one, I was not able to track it because it happens *AFTER* the expansion, and all UltiSnips python code... I even traced back to the vim function calling expand, and at the end of this function, the cursor is still at the right position, yet, it is moved once to the left after, and I have no f***ing idea how.

Reproduction step :

Use this branch :
https://github.com/hl037/ultisnips/tree/WTF

(it is the tiebreaker fix I planned to PR, but there is this bug...) The only thing I changed from the tiebreaker branch is the _VimCursor that has a +inf tie breaker to force it to the end (since it would be placed at the begin without).

Then, run : 

In terminal 2:
```
tmux -f /dev/null new -s vim bash --noprofile --norc
```

In terminal 1:
```
python ./test_all.py -v -i ZeroLength_Trivial
```

Now, type `test0`<kbd>Tab</kbd>

The first snippet is expanded correctly, go bac to insert mode and type `:message`

Here is an historic of the cursor position mutations (inside Ultisnips) and some log at some points (including the last [ ... , ... , ... , ... ] which is the from the end of the `UltiSnips#expand` function.

Start all over again but with the `test1`<kbd>Tab</kbd> snippet.

This time, the cursor ends up one column to the left from where it should be. Try again the `:message` command and you'll see the same history.

Yet, the cursor is still off...

hl037:
screen shots : 


# test0

![image](https://user-images.githubusercontent.com/2771764/140387880-80e76b24-e18c-4f1e-ae31-c5a40bb6a557.png)

`:messages` 
![image](https://user-images.githubusercontent.com/2771764/140387670-f859c2cd-1778-4db7-833c-ecc4723646e0.png)

# test1

![image](https://user-images.githubusercontent.com/2771764/140388008-966d38d2-d1c5-4522-97c3-77fa406474af.png)

`:messages`

![image](https://user-images.githubusercontent.com/2771764/140388116-d97e26d0-e837-495a-b4f5-29e8a80d59e7.png)

hl037:
Another thing I have observed, is that the bug happens only the FIRST time a snippet is expanded (then, it works as expected)

hl037:
Here are some log I recorded.

I launch once : 
```
[...]

PR #1408, opened 2021-11-08
[WIP] tiebreaker fix

hl037:
This is a work in progress.
This pr fixes most of the problems we get with the tie breaker. (yet, it may not be polished yet)
Unfortunately, some unit tests won't pass because of #1405.

But since I use this commit as start point for fixing #1405 by implementing the algorithm described in #1407, I think it deserves also a review =)

Basically, I reversed the logic of the "done" argument of the update method.

Then child of each parent are sorted lazily (first sorted at the end of the parsing, the add are perform as an insert-sort).

The only draw-back as it is, is that parent are tagged as "done" only on the next iteration, and that makes snippet recursion easily hitting the cyclic dependency check. To fix that, I'm thinking about either making the SnippetInstance not exposing its children and it's update function to do a similar algorithm as update_textobjects.

...But I prefer to first show this PR to get some reviews before pushing it further

I also cleaned and refactored the Position class, because some parts about the move/delta function pair were a bit obscure. Now, the delta is simply the difference between two positions. I also moved code constructing or dealing with delta from outside of the Position class to class methods of Position so that changing the internals of this class could all be made in only one file. That also ease tracking bugs or usage of the class by simple doing a usage search on these class methods.

These changes in the Position class make obsolete the unit tests (outside the tmux based test suite) because they were assessing the details of the class and not the actual public feature.
Thus, my confidence of these changes being right is based on the fact that all test cases which passed on the previous implementation still with this one (except as I said earlier the one or two that fail because of the tiebreaker fix and the selection/cursor getting out of sync)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,13 @@ In this shell we can then trigger the test execution:
     ... now inside container
     # ./test_all.py
 
+### Enable the remote debug server
+
+The test suite provides `--remote-pdb*` options equivalent to the config variables to enable the debug server during the test suite.
+Note that some tests may fail because the post-mortem will catch an expected exceptions and that these options are mainly useful for single test case debugging. 
+
+Check `./test_all.py --help` for more informations.
+
 ## Documenting
 
 User documentation goes into [`doc/UltiSnips.txt`](https://github.com/SirVer/ultisnips/blob/00_contributing/doc/UltiSnips.txt).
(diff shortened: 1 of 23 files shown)


hl037:
**Note** For reviewing, take only the last two commits since this PR is already based on #1404  : https://github.com/SirVer/ultisnips/pull/1408/commits/13dc0f9f13e7904837ec3a62a3c70948bcacb4fc

SirVer:
@hl037 I think you can merge master into this and we can start reviewing it. Wdyt?

SirVer:
Similar comment to https://github.com/SirVer/ultisnips/pull/1394#issuecomment-1148848489

PR #1416, opened 2021-12-18
Handle wildcards in runtimepath entries. Fixes #1414.

### LINKED ISSUE #1414: NVIM v0.6.0 take into account changes to the runtimepath
Snippets provided by plugins are no longer loaded using Neovim >= v0.6.0
This is the [commit](https://github.com/neovim/neovim/commit/9df7e022b498cb74ffbf5b8fd2ddc4dd5c04d127) introducing the change.
The paths of `start` packages in the `runtimepath` are no longer expanded, so instead of having an entry for each of them there is only one matching `/home/<username>/.local/share/nvim/site/pack/*/start/*`.

**Expected behavior:**
Load the snippets provided by plugins.

**Actual behavior:**
Snippets provided by plugins are not found.

**Steps to reproduce**
With the usual setup, open nvim using the correct version by using the appimage:
```sh
curl -LO https://github.com/neovim/neovim/releases/download/v0.6.0/nvim.appimage
chmod u+x nvim.appimage
```

Try to expand a snippet provided by a plugin stored in the `start` package's directory.

-----
I'm not comfortable with Python but I think it could be easily fixed in: https://github.com/SirVer/ultisnips/blob/2c83e40ce66814bf813457bb58ea96184ab9bb81/pythonx/UltiSnips/snippet/source/file/ulti_snips.py#L61-L64
My temporary fix is:
```py
            pth = normalize_file_path(
                os.path.expanduser(os.path.join(rtp, snippet_dir))
            )
            if '*' in pth:
                for fn in glob.glob(pth):
                    all_dirs.append(fn)
            else:
                all_dirs.append(pth)
```

-----

- **Operating System**: Arch Linux 5.15.7
- **Vim Version**: NVIM v0.6.0
- **UltiSnips Version**: 2c83e40ce66814bf813457bb58ea96184ab9bb81
- **Python inside Vim**: 3.9.9

smjonas:
Single wildcards ("*") are allowed in runtimepath entries, so expand them.
See `:h rtp<cr>71j` in Vim or `:h rtp-packages` in Neovim.

### DIFF
--- a/pythonx/UltiSnips/snippet/source/file/ulti_snips.py
+++ b/pythonx/UltiSnips/snippet/source/file/ulti_snips.py
@@ -61,7 +61,8 @@ def find_all_snippet_directories() -> List[str]:
             pth = normalize_file_path(
                 os.path.expanduser(os.path.join(rtp, snippet_dir))
             )
-            all_dirs.append(pth)
+            # Runtimepath entries may contain wildcards.
+            all_dirs.extend(glob.glob(pth))
     return all_dirs
 
 


SirVer:
Thanks @smjonas!

Issue #155, opened 2014-02-15
Saving visual selection gets stuck when selecting big texts


SirVer:
First of all, I'd like to say a big THANKS for this plugin that I discovered yesterday. It's really easy to use, documentation is precise and quite exhaustive, and the plugin itself is very powerful. Great job !

Well, my problem is that when selecting a quite big amount of text and pressing <tab> to reuse it in a snippet, vim gets stuck in the UltiSnips_SaveLastVisualSelection() function for a while. I don't exactly know what the _visual_content.conserve() function is doing, but an optimization would be welcomed :-).

My use-case is the following very simple C snippet, wrapping the selected text into a vim fold:

  snippet fold "Add a new vim fold" !b  
  /*{{{ ${1:topic} */                                                         

  ${0:${VISUAL}}                                                              

  /*}}} */  
  endsnippet

When working on a big C file (> 5k lines) without folds, and that I want to add some with this snippet... well I just can't :-D.

Launchpad Details: [#LP1178643](https://bugs.launchpad.net/bugs/1178643) LpZ_squall - 2013-05-10 13:46:39 +0200

kenng:
Is there a way for us to disable UltiSnips manually before inserting long line? Or let vim to disable UltiSnips when inserting long line? Probably disabling plugin is out of question in vim, but I just curious to know if it can.

seletskiy:
@kenng: The only workaround so far is to jump out of snippet (e.g. hit `<C-J>` until you will encounter last placeholder) and then paste text wherever you want.

mgedmin:
I have some suggestions maybe?

- could you use [difflib](https://docs.python.org/3/library/difflib.html) instead of implementing a manual diff algorithm by hand?
- could you abort snippet expansion logic if the edit is too large according to some criterion (simple length, or `cost` reaching a high value, or simply `time.time()` reaching some timeout value since the beginning of the computation)?

mgedmin:
BTW Vim 8.1.1321 added the [`listener_add()`](https://vimhelp.org/eval.txt.html#listener_add%28%29) API that lets you register a callback and be informed about every change the user makes.

SirVer:
@mgedmin This looks perfect, I believe this can be used to solve this problem once and for all.

SirVer:
I looked into using `listener_add`, and unfortunately that API is not exposing all information necessary for UltiSnips. It misses what edits have been taken place (i.e. if there are deletions insertions or changes), so this will not help us here. 

[...]

PR #1417, opened 2021-12-22
Don't trigger _leaving_buffer for floating windows

dhleong:
This fixes an issue where snippets will not be navigable when floating
windows are used to present extra information, such as with nvim-cmp
when "accepting" a completion triggers a snippet expansion.

This PR only handles neovim floating windows; I'm not sure whether vim8 popup windows require similar consideration.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -33,7 +33,7 @@ and set of keystrokes that reproduces your problem.
 - **Vim Version**: <!-- first two lines of `:version` output -->
 - **UltiSnips Version**: <!-- e.g. 3.1. If you're using version from git 
                               run: `git rev-parse origin/master` -->
-- **Python inside Vim**: <!-- e.g. 2.7.14 / 3.6.5. If unsure run inside vim:
+- **Python inside Vim**: <!-- e.g. 3.9.1 / 3.6.5. If unsure run inside vim:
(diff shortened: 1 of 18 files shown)


SirVer:
This might be a bad idea....

Whenever UltiSnips sees a cursor movement it will try to figure out if this cursor movement has changed the buffer in some way and update its internal state representation. If the change is large, this might take a long time, since this is essentially a full diff on the buffer in the worst case. The reason we terminate snippets when leaving the current buffer is that this diffing between an existing and a new buffer will likely be so expensive that Vim will not react anymore. Examples of this issue are #155 and #1074, but this is among the most reported issues with UltiSnips.

What you do here is circumvent this escape hatch which will likely mean that this bug triggers much more often. You should be able to figure this out by launching many nested snippets that are all active and have lots of text, then trigger the floating window with a lot of text.

SirVer:
@dhleong I took a look at this again and my last comment still stands, however I am willing to merge this until somebody complains, maybe I am wrong about how bad the problem is in the real world.

This however fails on 7.4 since `win_getid` is not defined. I feel it is no longer necessary to support Vim 7.4, so I'll remove its support from master.

SirVer:
Okay, let's have it and see if something blows up in the near future down the line. Thanks for the contribution @dhleong!

PR #1421, opened 2022-01-16
precompile python code in snippets

ces42:
This commit changes the classes `SnippetDefinition` and `PythonCode` so that the code contained in Python interpolation, global snippets, snippet contexts and snippet actions is compiled (using Python's builtin `compile`) when an instance is created, instead of having `exec` compile the code every time it is executed.
This is (supposed to be) a non-functional change.

Compile times for Python code are very slow, e.g. compiling 63 lines of global snippets in my latex snippets file takes about 450 µs on my machine while executing the compiled code only takes 1.3 µs.

### DIFF
--- a/pythonx/UltiSnips/snippet/definition/base.py
+++ b/pythonx/UltiSnips/snippet/definition/base.py
@@ -500,6 +500,7 @@ class SnippetDefinition:
             last_re=self._last_re,
             globals=self._globals,
             context=self._context,
+            _compiled_globals = self._compiled_globals,
         )
         self.instantiate(snippet_instance, initial_text, indent)
         snippet_instance.replace_initial_text(vim_helper.buf)
--- a/pythonx/UltiSnips/snippet/definition/base.py
+++ b/pythonx/UltiSnips/snippet/definition/base.py
@@ -111,7 +111,7 @@ class SnippetDefinition:
         self._globals = globals
         self._compiled_globals = compile("\n".join([
                     "import re, os, vim, string, random",
-                    "\n".join(self._globals.get("!p", [])).replace("\r\n", "\n")
+                    "\n".join(globals.get("!p", [])).replace("\r\n", "\n")
                 ]), '<global-snippets>', 'exec')
         self._location = location
         self._context_code = context
--- a/pythonx/UltiSnips/snippet/definition/base.py
+++ b/pythonx/UltiSnips/snippet/definition/base.py
@@ -16,6 +16,11 @@ from UltiSnips.text import escape
 from UltiSnips.text_objects import SnippetInstance
 from UltiSnips.text_objects.python_code import SnippetUtilForAction
 
+# We'll end up compiling the global snippets for every snippet so
+# caching compile() shoud pay off
+from functools import lru_cache
+compile = lru_cache(compile)
+
 __WHITESPACE_SPLIT = re.compile(r"\s")
 
 
@@ -104,10 +109,18 @@ class SnippetDefinition:
         self._matched = ""
         self._last_re = None
         self._globals = globals
+        self._compiled_globals = compile("\n".join([
+                    "import re, os, vim, string, random",
+                    "\n".join(self._globals.get("!p", [])).replace("\r\n", "\n")
+                ]), '<global-snippets>', 'exec')
         self._location = location
         self._context_code = context
+        if context:
+            self._compiled_context_code = compile('snip.context = ' + context, '<context-code>', 'exec')
         self._context = None
         self._actions = actions or {}
+        self._compiled_actions = {action: compile(source, '<action-code>', 'exec') for action,
+                source in self._actions.items()}
 
         # Make sure that we actually match our trigger in case we are
         # immediately expanded.
@@ -154,17 +167,9 @@ class SnippetDefinition:
             locals["visual_text"] = visual_content.text
             locals["last_placeholder"] = visual_content.placeholder
 
-        return self._eval_code("snip.context = " + self._context_code, locals).context
+        return self._eval_code("snip.context = " + self._context_code, locals, self._compiled_context_code).context
 
-    def _eval_code(self, code, additional_locals={}):
-        code = "\n".join(
-            [
-                "import re, os, vim, string, random",
-                "\n".join(self._globals.get("!p", [])).replace("\r\n", "\n"),
-                code,
-            ]
-        )
-
+    def _eval_code(self, code, additional_locals={}, compiled_code=None):
         current = vim.current
 
         locals = {
@@ -180,14 +185,23 @@ class SnippetDefinition:
         snip = SnippetUtilForAction(locals)
 
         try:
-            exec(code, {"snip": snip})
+            glob = {"snip": snip}
+            exec(self._compiled_globals, glob)
+            exec(compiled_code if compiled_code is not None else code, glob)
         except Exception as e:
+            code = "\n".join(
+                [
+                    "import re, os, vim, string, random",
+                    "\n".join(self._globals.get("!p", [])).replace("\r\n", "\n"),
+                    code,
+                ]
+            )
             self._make_debug_exception(e, code)
             raise
 
         return snip
 
-    def _execute_action(self, action, context, additional_locals={}):
+    def _execute_action(self, action, context, additional_locals={}, compiled_action=None):
         mark_to_use = "`"
         with vim_helper.save_mark(mark_to_use):
             vim_helper.set_mark_from_pos(mark_to_use, vim_helper.get_cursor_pos())
@@ -198,7 +212,7 @@ class SnippetDefinition:
 
             locals.update(additional_locals)
 
-            snip = self._eval_code(action, locals)
+            snip = self._eval_code(action, locals, compiled_action)
 
             if snip.cursor.is_set():
                 vim_helper.buf.cursor = Position(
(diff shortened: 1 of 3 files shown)


pythonx/UltiSnips/snippet/definition/base.py
```diff
@@ -16,6 +16,11 @@
 from UltiSnips.text_objects import SnippetInstance
 from UltiSnips.text_objects.python_code import SnippetUtilForAction
 
+# We'll end up compiling the global snippets for every snippet so
+# caching compile() shoud pay off
```
SirVer, line 20: typo: `should`

pythonx/UltiSnips/snippet/definition/base.py
```diff
@@ -16,6 +16,11 @@
...
+from functools import lru_cache
+compile = lru_cache(compile)
```
SirVer, line 22: Please do not alias. Instead make an wrapper function: `lru_cached_compile` so the code becomes unambiguous.

pythonx/UltiSnips/text_objects/python_code.py
```diff
@@ -12,6 +12,11 @@
 from UltiSnips.vim_state import _Placeholder
 import UltiSnips.snippet_manager
 
+# We'll end up compiling the global snippets for every snippet so
```
SirVer, line 15: same comment.

SirVer:
Thanks for this contribution! 

Did you actually time this?, my understanding is that python does cache compiles internally already, similar to how `import re` keeps the latest regex around for caching. So my assumption is that only the first execution is slow, after it should be fast.

Assuming your data finds that you are correct and it compiles every time. Now this opens up another question: Is it better to be slow at snippet execution or slow at snippet parsing (your new change). This is of course a philosophical question and opinions might vary, but have you thought about this?

ces42:
`compile()` does not cache anything. Just run the following code in the REPL and see how long the loops take:
```python
s = """
def fun():
    return True
a = fun()
"""

for i in range(2*10**5):
    exec(s)

cmp = compile(s, '<s>', 'exec')
for i in range(2*10**5):
    exec(cmp)
```
I did not actually time this because I don't know how to. I tried to write a vimscript loop calling `feedkeys()` in insert mode that should trigger a snippet with a context condition, but this does not make UltiSnips behave as expected. If I understand correctly this is due to the overhead of python calls being too large to keep up with vim's loop (I also did most of the testing on neovim which should exacerbate the problem).

I was hoping someone could suggest a method for timing this. If there was some way to see logging or stdout from the pyth

ces42:
`compile()` does not cache anything. Just run the following code in the REPL and see how long the loops take:
```python
s = """
def fun():
    return True
a = fun()
"""

for i in range(2*10**5):
    exec(s)

cmp = compile(s, '<s>', 'exec')
for i in range(2*10**5):
    exec(cmp)
```
I did not actually time this because I don't know how to. I tried to write a vimscript loop calling `feedkeys()` in insert mode that should trigger a snippet with a context condition, but this does not make UltiSnips behave as expected. If I understand correctly this is due to the overhead of python calls being too large to keep up with vim's loop (I also did most of the testing on neovim which should exacerbate the problem).

I was hoping someone could suggest a method for timing this. Having some way to see logging or stdout from the python files in UltiSnips (preferably in real time) would probably be sufficient.

The answer to @SirVer's last question, seems very straightforward to me. Compile times should be a couple ms at most even for large files which probably isn't even noticeable once at startup but they could be meaningful if they happen every time a snippet is triggered (or maybe even not triggered, in the case of `context`). Furthermore, if I understand correctly, the code inside a Python interpolation is executed **at every keypress** when edition a tabstop. This means that all global snippets have to be compiled once for every python interpolation in the snippet I'm currently expanding, which seems like a huge waste of resources.

SirVer:
> I was hoping someone could suggest a method for timing this. Having some way to see logging or stdout from the python files in UltiSnips (preferably in real time) would probably be sufficient.

I used dump printf debugging into a file using `pythonx/UltiSnips/debug.py:debug` which might work.

I played around with your example with `ipythons` `%time` and `%timeit` methods and you are definitely right that there is no caching going on. Hence I follow your arguments about this being a strict improvement in most use cases and am very excited about accepting your patch! Would you address the few comments, then I'll merge this?

ces42:
I managed to do some proper timing of this. The relevant code is in the `profiling` branch in my repository. Running `test_all.py ContextSnippets_LotsOfGlobals` will try to expand a context snippet with a lots of global snippets 500 times and the profiling info will be sent to the dump file.
Currently I don't observe any speedup with my changes so more investigation is necessary before this can be merged.

hl037:
I don't have really time to check it but what does your profiling tells you ? ...I would actually think the communication with vim is dominating the timing IMHO.

ces42:
@SirVer I think the profiling I wrote should be accurate. I'm just using the testing framework as a convenient way of making vim try to expand a snippet a lot of times. The actual profiling is done all inside Python, see [https://github.com/ces42/ultisnips/commit/1cd745ab9e16025cfe0facafc2ea8069a3548a37](here).

Compiling/executing the global snippets in the test takes ~500 us on my machine.
The logging in `/tmp/file.txt` looks like this without the PR (this is on vim, not neovim)
```
1 calls to _try_expand have taken  2.9ms
2 calls to _try_expand have taken  3.2ms
3 calls to _try_expand have taken  3.7ms
4 calls to _try_expand have taken  4.1ms
[...]
1497 calls to _try_expand have taken  1039.4ms
1498 calls to _try_expand have taken  1040.6ms
1499 calls to _try_expand have taken  1041.5ms
1500 calls to _try_expand have taken  1042.1ms
```
so the average time spent checking whether a snippet can be expanded, including executing the global snippets + context condition seems to be .7ms. With the PR I get
```
1 calls to _try_expand have taken  1.8ms
2 calls to _try_expand have taken  2.1ms
3 calls to _try_expand have taken  2.5ms
4 calls to _try_expand have taken  2.8ms
[...]
1497 calls to _try_expand have taken  949.8ms
1498 calls to _try_expand have taken  950.6ms
1499 calls to _try_expand have taken  951.3ms
1500 calls to _try_expand have taken  951.8ms
```
so the average time is ~.63ms. The expected speedup of .5ms does not happen. Actually looking at the first lines of the logs is making me believe that maybe @SirVer was right and there actually is some caching going on?? Because the first call **is** significantly faster using the PR (this has been a consistent patter across multiple runs).

SirVer:
@ces42 I understand better now what you did and indeed find it puzzling too. Caching should yield inverse results, right? I.e. the first time should be slower than all consecutive times, but this is not what we see: it is indeed faster the first time around.

 What is your opinion on how to proceed with this PR?

ces42:
@SirVer Hi, I forgot to write I was doing the test wrong (in several different ways I think). Doing things correctly I get the following result w/o compiling
```
1 calls to _try_expand have taken  1.9ms
2 calls to _try_expand have taken  2.9ms
3 calls to _try_expand have taken  3.1ms
4 calls to _try_expand have taken  4.1ms
[...]
1495 calls to _try_expand have taken  802.9ms
1496 calls to _try_expand have taken  803.1ms
1497 calls to _try_expand have taken  804.0ms
1498 calls to _try_expand have taken  804.3ms
```
and with compiling
```
1 calls to _try_expand have taken  2.7ms
2 calls to _try_expand have taken  3.1ms
3 calls to _try_expand have taken  3.3ms
4 calls to _try_expand have taken  3.5ms
[...]
1497 calls to _try_expand have taken  414.4ms
1498 calls to _try_expand have taken  414.7ms
1499 calls to _try_expand have taken  415.0ms
1500 calls to _try_expand have taken  415.2ms
```
The speedup is not quite as large as expected (400us * 1500 would be 600ms) but it's definitely there.

ces42:
I haven't explicitly timed the code doing python interpolation but I'm quite confident that there should be a similar speedup in that situation.

SirVer:
@ces42 I fixed a bug with old python version and merged master that had CI fixed. This is green, so let's have it. Big Thanks for contributing! ❤️

SirVer:
@ces42 I had to roll this back since it triggered #1428 and I had no time to dig deeper why it happened yet. I would appreciate if you could have a look if your time allows, since the feature is of course very useful!

ces42:
Oops, sorry for introducing a bug (I though I had checked whether the relative order of global snippets and snippets using the mattered).

In theory this shouldn't be hard to fix, although the iterative nature in which snippet files are read (using a bunch of generators: `_parse_snippets_file()` and `UltiSnipsFileSource._parse_snippets_file()` in `snippet/source/file/ulti_snips.py` which are iterated over in `SnippetFileSource._parse_snippets()` in `snippet/source/file/base.py`) might make things a bit more complicated.
What is the reason for reading and defining snippets in an iterator instead of parsing the all in a loop and returning a list of snippets? Is it so that if there is a syntax error at some point in the file the snippets before that point can still be salvaged?

I'll be able to have a better look at this over the weekend.

SirVer:
@ces42 No worries about introducing a bug. This happens when new features are added. If you fix the problem, please add a test case and we can right away merge your fix back in.

ces42:
What is the purpose of https://github.com/SirVer/ultisnips/blob/c428cdf605658004ae067e862ca7ec3f3e22799d/pythonx/UltiSnips/snippet/definition/base.py#L112-L114 ?

Should this be an assert? In any case, I think this introduces a bug similar to #1428 (but unrealted to the precompilation) because it might cause UltiSnips to try to evaluate a context condition that might only be defined later on in the file.

PR #1423, opened 2022-01-17
Remove Python 3.5 and add 3.10.

SirVer:
Official 3.5 support has ended in January and Python 3.10 is a thing now.


hl037:
I think It can only makes things better !
Sorry Holger for not having sent my mail setting up a meeting to change the tab navigation algorithm, I got covid, then was really really busy, I hope to have soon some time (not easy since I'm also preparing the French Computer science aggregation in the hope to become a teacher...)

SirVer:
@hl037 No worries, Leó. I am not going anywhere, just reach out when you want to chat :).

PR #1426, opened 2022-01-22
Expose match object to python code in context conditions

ces42:
As the title says. Actually more because `match` will also be available inside the scope of snippet actions.

I implemented it so that it is the same as the `match` variable exposed to the code in python interpolation. (There is a tiny contradiction to the documentation here, also for the match variable in python interpolation. When a snippet is not re triggered, `match` is `None` and not undefined as the docs suggest. This is probably fine.)

I can also add an example use case to the documentation.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1432,7 +1432,7 @@ In that case snippet should be defined using this syntax: >
 
     snippet trigger_word "description" "expression" options
 
-The context can be defined using a special header: >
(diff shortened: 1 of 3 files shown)


SirVer:
@ces42 Can you add a test for this?

ces42:
@SirVer I added a test. Do you think this covers it?

SirVer:
Yes, absolutely. Thanks!

PR #1429, opened 2022-01-25
document that regex snippets are expanded in-word

### LINKED ISSUE #1420: Snippet option `r` seems to imply option `i`
**Expected behavior:**
If I define a snippet
```
snippet "abc" r
abcdefghijkl...
endsnippet
```
I would expect this to be equialent to the snippet defined with `snippet abc`.

**Actual behavior:**
The first snippet will trigger on `sabc` while (in-word) the second one won't. It seems that option `r` automatically implies option `i`

**Steps to reproduce**
as described above

This is not a terrible thing because if I'm already defining a regex trigger I can just do the word-checking myself. But it would be good to have this described in the documentation

-----
- **Operating System**: Ubuntu 21.10
- **Vim Version**: neovim 0.6.1
- **UltiSnips Version**: current (`2c83e40ce66814bf813457bb58ea96184ab9bb81`)
- **Python inside Vim**: 3.9.7

ces42:
Also document that option `w` overrides option `i`. Fixes #1420.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -684,7 +684,7 @@ The options currently supported are: >
        preceded and followed by non-word characters. Word characters are
        defined by the 'iskeyword' setting. Use this option, for example, to
        permit expansion where the tab trigger follows punctuation without
-       expanding suffixes of larger words.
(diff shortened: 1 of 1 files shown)


SirVer:
Great! Thank you.

PR #1432, opened 2022-01-30
expose text in line before snippet to context conditions (faster on neovim)

ces42:
This would add a `before` attribute to the `snip` object in context conditions. `snip.before` should be equivalent to `snip.buffer[snip.line][:snip.column+1]`. **EDIT:** Actually, it's not because `vim.current.window.cursor` and by extension `snip.column` count bytes, not characters, see https://github.com/neovim/pynvim/issues/383.

However this implementation is faster on neovim (by ~ 0.7 ms) because it avoids an api call and the associated IPC overhead. An additional advantage is that it is more concise and avoids confusion about 1- and 0-based indexing.

If you decide you want to merge this I can write documentation for it.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -1466,6 +1466,12 @@ Global variable `snip` will be available with following properties:
         - 'current_text' - text in the placeholder on the moment of selection;
         - 'start' - placeholder start on the moment of selection;
         - 'end' - placeholder end on the moment of selection;
+    'snip.before' - contains the text in the current line up to and including
+        the matched snippet. Since 'snip.column' counts bytes, not characters,
+        this is **not** equivalent to 'snip.buffer[snip.line][:snip.column+1]'.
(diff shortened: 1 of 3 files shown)


SirVer:
So I hate the argument that this is good because it is faster on Neovim.... Working around the design flaws of NeoVim is not really the job of plugins I think. 

I like the convenience argument much better. Let's have it for this reason! Could you add a test + documentation?

SirVer:
@ces42, friendly ping on this.

ces42:
Added doc and test. Sorry this took so long.

Also I found an even better reason to add this: `vim.current.window.cursor` and by extension `snip.column` count bytes, not characters, see https://github.com/neovim/pynvim/issues/383. so using `snip.buffer[snip.line][:snip.column+1]` can lead to all sorts of fun bugs for people using non-ascii characters.

SirVer:
@ces42 Thanks so much!

Issue #1364, opened 2021-06-08
UltiSnipsEdit opens blank files instead of existing snippets


idbrii:
**Expected behavior:**

`:UltiSnipsEdit` always opens an existing file if one can be found.

**Actual behavior:**

`:UltiSnipsEdit` creates ~/.vim/ultisnips and opens an empty file.

(And the folder name doesn't even match the case of `g:UltiSnipsSnippetDirectories`.)

**Steps to reproduce**

1. Create ~/vimfiles (if it already exists, rename it and create a new one). (This step prevents the "not able to find a default directory" error.)
1. Using minimal vimrc run `gvim -Nu ~/.vim/reprovimrc.vim -U NONE`
2. Type `if<Tab>` to show that snippets are found and work.
2. `:UltiSnipsEdit`
1. Bug: An empty file is opened.
2. `:UltiSnipsEdit!`
1. Bug?: The output includes two files that don't exist (3,4):

```
* 1: c:\users\idbrii\.vim\bundle\snippets\ultisnips\all.snippets
* 2: c:\users\idbrii\.vim\bundle\snippets\ultisnips\snippets.snippets
  3: c:\users\idbrii\vimfiles\ultisnips\all.snippets
  4: c:\users\idbrii\vimfiles\ultisnips\snippets.snippets
Type number and <Enter> or click with the mouse (q or empty cancels): 
```

## minimal vimrc 
```vim
" all plugins should work with sensible as a baseline.
let s:plugins = ['sensible']

let s:plugins += ["ultisnips"]
" Some existing snippets to open.
let s:plugins += ["snippets"]

set runtimepath-=~/.vim
set runtimepath-=~/.vim/after
for plugin in s:plugins
    exec "set runtimepath^=~/.vim/bundle/". plugin
    exec "set runtimepath+=~/.vim/bundle/". plugin ."/after"
endfor
set viminfofile=NONE
colorscheme desert

let $PYTHONHOME = expand('$USERPROFILE/scoop/apps/python39/current')
let $PYTHONPATH = $PYTHONHOME . "/Lib"
let $PATH = $PATH . ';' . $PYTHONHOME
```

-----

- **Operating System**: Windows 10
- **Vim Version**: 
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Feb 10 2021 23:02:05)
MS-Windows 64-bit GUI version with OLE support
- **UltiSnips Version**: 
[...]

PR #1440, opened 2022-02-14
Resolve paths to avoid duplicates in UltiSnipsEdit

idbrii:
Use realpath to get the actual path so symlink directories are counted
as equal and excluded from the set of output.

When dotfiles are symlinked:

> dir c:\users\idbrii\.vim
...
2020-12-08  04:01 PM    <SYMLINKD>     .vim [C:\dotfiles\vim]

`:UltiSnipsEdit!` outputs:
```
    1: C:\Users\idbrii/.vim/bundle/david-snippets/UltiSnips/fugitive.snippets
  * 2: c:\dotfiles\vim\bundle\david-snippets\ultisnips\all.snippets
  * 3: c:\dotfiles\vim\bundle\david-snippets\ultisnips\all\shebang.snippets
  * 4: c:\dotfiles\vim\bundle\snippets\ultisnips\all.snippets
  * 5: c:\dotfiles\vim\bundle\project\ultisnips\all.snippets
  * 6: c:\users\idbrii\.vim\bundle\david-snippets\ultisnips\all.snippets
  * 7: c:\users\idbrii\.vim\bundle\david-snippets\ultisnips\all\shebang.snippets
    8: c:\users\idbrii\.vim\ultisnips\all.snippets
    9: c:\users\idbrii\.vim\ultisnips\fugitive.snippets
```
2,6 and 3,7 are duplicates.

This fixes the same issue on linux with similar symlink setup:
```
  $ ls -l ~/.vim
  lrwxrwxrwx 1 idbrii idbrii 49 Jan 24  2021 /home/idbrii/.vim -> /home/idbrii/dotfiles/vim
```

### DIFF
--- a/pythonx/UltiSnips/snippet/source/file/common.py
+++ b/pythonx/UltiSnips/snippet/source/file/common.py
@@ -8,6 +8,7 @@ import os.path
 
 def normalize_file_path(path: str) -> str:
     """Calls normpath and normcase on path"""
+    path = os.path.realpath(path)
     return os.path.normcase(os.path.normpath(path))
 
 
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -23,6 +23,9 @@ from UltiSnips.snippet.source import (
     find_all_snippet_files,
     find_snippet_files,
 )
+from UltiSnips.snippet.source.file.common import (
+    normalize_file_path,
+)
 from UltiSnips.text import escape
 from UltiSnips.vim_state import VimState, VisualContentPreserver
 from UltiSnips.buffer_proxy import use_proxy_buffer, suspend_proxy_edits
@@ -97,7 +100,9 @@ def _get_potential_snippet_filenames_to_edit(
         potentials.update(ft_snippets_files)
         if not ft_snippets_files:
             # If there is no snippet file yet, we just default to `ft.snippets`.
-            potentials.add(os.path.join(snippet_dir, ft + ".snippets"))
+            fpath = os.path.join(snippet_dir, ft + ".snippets")
+            fpath = normalize_file_path(fpath)
+            potentials.add(fpath)
     return potentials
 
 


dlukes:
Just FYI, this can break `UltiSnipsEdit` for symlinked snippet directories. My personal snippet dir is actually under `~/.files/snippets`, but I symlink it to `~/.config/nvim/UltiSnips` (for vanilla Vim, this would be `~/.vim/UltiSnips`). After this PR, it doesn't get added to the set of `potentials` as `~/.config/nvim/UltiSnips`, but in its resolved form `~/.files/snippets`. Which means it then gets skipped because of [this check](https://github.com/SirVer/ultisnips/blob/f5ccf0977c611ffd774ca180774959301baaffad/pythonx/UltiSnips/snippet_manager.py#L893), as it seems to reside outside of the user's config directory.

In case anyone else comes across this, the workaround is explicitly setting a `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit`, which is probably a good idea anyway as it avoids having to rely on guesswork to determine it.

idbrii:
[Relevant code](https://github.com/SirVer/ultisnips/blob/f5ccf0977c611ffd774ca180774959301baaffad/pythonx/UltiSnips/snippet_manager.py#L893):

```python
        if (len(all_snippet_directories) != 1 and not has_storage_dir) or (
            has_storage_dir and bang
        ):
            # Likely the array contains things like ["UltiSnips",
            # "mycoolsnippets"] There is no more obvious way to edit than in
            # the users vim config directory.
            dot_vim_dir = vim_helper.get_dot_vim()
            for snippet_dir in all_snippet_directories:
                if Path(dot_vim_dir) != Path(snippet_dir).parent:
                    continue
                potentials.update(
                    _get_potential_snippet_filenames_to_edit(snippet_dir, filetypes)
                )
```

That comment seems like it might be outdated so maybe the check is no longer needed. Was introduced in 298f4c255dd7048a99d968d91fc001b6e2203e65, but at first glance I don't understand its purpose.

[find_all_snippet_directories is building full paths](https://github.com/SirVer/ultisnips/blob/f5ccf0977c611ffd774ca180774959301baaffad/pythonx/UltiSnips/snippet/source/file/ulti_snips.py#L61-L63) (since rtp has absolute paths), so `all_snippet_directories` wouldn't contain things like `["UltiSnips", "mycoolsnippets"]`. Although, if we expected folder names, I'm not sure that check would work unless our cwd happened to be .vim/

dlukes:
> at first glance I don't understand its purpose

In my case, `all_snippet_directories` ends up containing two absolute paths, one to my private snippet dir and another to a snippet dir provided by a plugin I use. So the check is meaningful, it's intended to identify the former.

Unfortunately, it breaks when the private snippet dir is a symlink and it gets resolved, and I don't really see a way of fixing that, short of reverting the resolving. One just has to explicitly set `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit`.

dlukes:
FWIW though, is the resolving really necessary? I just reverted to the commit prior to this change (3f8bdedd) and made my config dir a symlink (`/Users/david/.config/nvim -> nvim-real/`), and I don't see any duplicate entries when I run `:UltiSnipsEdit!`.

It looks like resolving paths on UltiSnips' side just papers over the root cause, which is duplicate entries on the runtimepath? And that's something the user should probably fix on their side.

AFAICS, with vanilla Vim and a symlinked (empty) `/Users/david/.vim -> .files/vim`, I don't get any duplicate entries on the runtimepath:

```
runtimepath=~/.vim,/usr/share/vim/vimfiles,/usr/share/vim/vim82,/usr/share/vim/vimfiles/after,~/.vim/after
```

So if you have both `c:\dotfiles\vim` and `c:\users\idbrii\.vim` on your runtimepath, which are effectively the same dir, it's probably something you can prevent from happening in your config and the duplicate snippet entries would go away? Just guessing though.

SirVer:
@idbrii @dlukes I feel like I went back and forth on this topic (i.e. how to deal with symlinks) before and I vaguely remember similar discussions to the one we currently have. I feel I landed at the same resolve than @dlukes did, i.e. that it is not correct to resolve symlinks. I did just not remember when I reviewed and merged this pull request.

@idbrii can you outline why you think this change should stay in and not be reverted? I.e. what problem did it solve for you, let's analyse it and figure out if another solution would be more correct.

idbrii:
The issue was not with snippets in .vim, but snippets in plugins. Snipped from OP:

> `:UltiSnipsEdit!` outputs:

```
  * 2:      c:\dotfiles\vim\bundle\david-snippets\ultisnips\all.snippets
  * 6: c:\users\idbrii\.vim\bundle\david-snippets\ultisnips\all.snippets

  * 3:      c:\dotfiles\vim\bundle\david-snippets\ultisnips\all\shebang.snippets
  * 7: c:\users\idbrii\.vim\bundle\david-snippets\ultisnips\all\shebang.snippets
```

c:\dotfiles\vim\bundle\david-snippets only appears once in my 'runtimepath' (I'm using pathogen):

	:echo &rtp->split(',')->filter( {k,v -> stridx(v, "david-snippets") >= 0})
	['c:\dotfiles\vim\bundle\david-snippets']

@dlukes Can you try your repro with snippet-providing plugins in the rtp? Do `:UltiSnipsEdit` and `:UltiSnipsEdit!` both behave as documented?


> In my case, all_snippet_directories ends up containing two absolute paths, one to my private snippet dir and another to a snippet dir provided by a plugin I use. So the check is meaningful, it's intended to identify the former.

Sounds like you don't want snippets provided by plugins to be included. That may be the correct behaviour for `:UltiSnipsEdit`. I'm not sure what "private" means in this context:

> The UltiSnipsEdit command opens a private snippet definition file for the current filetype. If no snippet file exists, a new file is created. 

However, my snippets are in (my own) plugins, so I do want to see them. I'm using `:UltiSnipsEdit!` (bang) which is supposed to list *all* snippet directories:

> If used as UltiSnipsEdit! all public snippet files that exist are taken into account too. 


I can't see why anyone would want the same folder printed twice regardless of how they're using snippets. I also can't see a reason to exclude folders outside of ~/.vim when listing all snippets.

However, if  `:UltiSnipsEdit`'s "private" snippets means ones in `~/.vim` (or $VIM_HOME), then always resolving is likely incorrect and we need to uniquify without modifying the paths.

dlukes:
I did some more testing with the commit just prior to this PR.

Without setting `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit`, I don't have any duplicate entries in the list generated by `:UltiSnipsEdit!`, as [I previously indicated](https://github.com/SirVer/ultisnips/pull/1440#issuecomment-1109240423).

But when I *do* set it, I *do* get duplicate entries in the list, even though I have no duplicates in my runtimepath -- like @idbrii. So there is/was something funky going on in the snippet detection logic.

I'm a bit too tired to start digging into it right now, as the conditions are fairly involved and I'm not familiar with the codebase, but in general, could we do something along the following lines:

1. For `:UltiSnipsEdit`, symlinks should not be resolved, because the private snippet file to open is determined as the one which is under the `dot_vim_dir` prefix. Duplicates in the intermediate list of potential candidates don't matter in this case, as only one result will ultimately be kept anyway.
2. For `:UltiSnipsEdit!`, the reverse is true. We don't care about the `dot_vim_dir` prefix, we show all found snippet files. So we might as well resolve symlinks to get rid of duplicates.

Only drawback I can think of right now (4AM): the same snippet file will be shown under different paths when using `:UltiSnipsEdit` (symlinked) vs. `:UltiSnipsEdit!` (resolved). But in the first case, you don't really care about the path anyway since the file gets directly opened, so it doesn't seem like too much of a problem in practice.

@idbrii If you have a bit of free time, could you maybe whittle down your config into a minimal reproducible example, so that we can all reliably test against the same setup and agree on what we're seeing?

dlukes:
Ignore what I said in the previous post, I think I have a clearer understanding of the issue now. I think the problem is that the snippet search path is basically `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` + runtimepath, so even if the runtimepath is alias-free and there's nothing wrong with it, adding `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` into the mix can introduce aliasing.

@idbrii Are you by any chance setting `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` to `c:\dotfiles\vim\bundle\david-snippets\ultisnips`? If so, could you try setting it to `c:\users\idbrii\.vim\bundle\david-snippets\ultisnips` instead?

In my case, the reason for duplicate entries was analogous to this. `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` was a directory outside of the runtimepath, but there was a symlink on the runtimepath which also pointed to it (because obviously, I don't want to only edit the snippets, I also want to use them). Hence duplicates. Setting `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` to the equivalent runtimepath symlink instead got rid of them.

Personally, I think telling people who mind having duplicates on the `UltiSnipsEdit!` list that they should avoid having `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` be an alias to a directory on the runtimepath is a defensible position. Alternatively, if we'd rather allow this while still avoiding duplicates and showing only the alias under the `g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit` prefix, something like https://github.com/dlukes/ultisnips/commit/e3ff99d2d28f07a10dddae5a292414ba771e4a0d could be done. (That branch already contains a revert of this PR as the previous commit.)

idbrii:
I'm using this:
```
let g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit = '~/.vim/bundle/david-snippets/UltiSnips/'
```

Which [gets expanded](https://github.com/SirVer/ultisnips/blob/master/pythonx/UltiSnips/snippet_manager.py#L872) but not resolved.

On 3f8bded, with that setting, I wasn't able to repro my issue on a stripped down setup (gvim -Nu ~/.vim/reprovimrc.vim where reprovimrc sets rtp to just my snippets). However, I still see it with my full config.

I can confirm that on 3f8bded doing this gives me duplicated entries my stripped down setup:
```
let g:UltiSnipsSnippetStorageDirectoryForUltiSnipsEdit = 'c:/dotfiles/vim/bundle/david-snippets/ultisnips'
```

----

I'm fine with rolling back my change if you'd like to. Resolving paths seems like the right thing to me, but I remember now that `:UltiSnipsEdit` isn't quite the command I expect (#1364) so my intuition is likely wrong. Generally, my desire is to open an existing snippet file (always in a plugin) and see all snippet files (incl. plugins). I was a bit confused when I started testing this because at some point I've overridden `:UltiSnipsEdit` with a wrapper around `:Vedit` from [vim-scriptease](https://github.com/tpope/vim-scriptease).

One final thought: would it make sense to resolve all paths **including the ~/.vim** path for checking "is this a private snippet"? That way we're always comparing resolved paths.

dlukes:
> One final thought: would it make sense to resolve all paths including the ~/.vim path for checking "is this a private snippet"? That way we're always comparing resolved paths.

I think that doesn't always work. It works when the entire `.vim` directory is a symlink:

```
.
├── .vim -> vim-real
└── vim-real
   └── UltiSnips
```

Then, `.vim` resolves to `.../vim-real` and `UltiSnips` to `.../vim-real/UltiSnips`, so the former is the parent of the latter, [which is the check used to determine the private snippet dir](https://github.com/SirVer/ultisnips/blob/f5ccf0977c611ffd774ca180774959301baaffad/pythonx/UltiSnips/snippet_manager.py#L893) if no storage dir is set.

But it doesn't work when `UltiSnips` itself is a symlink to somewhere else:

```
.
├── .vim
│  └── UltiSnips -> ../UltiSnips-real
└── UltiSnips-real
```

In this case, `.vim` resolves to `.../.vim`, and `UltiSnips` to `.../UltiSnips-real`, so the former is no longer the parent of the latter.

> However, I still see it with my full config.

If you're setting `~/.vim/bundle/david-snippets/UltiSnips/` as your storage dir and have `c:\dotfiles\vim\bundle\david-snippets` on your runtimepath (as per one of your previous comments), I think that makes sense. So then using `c:\dotfiles\vim\bundle\david-snippets\UltiSnips` as your storage dir (to make it match what's on the runtimepath) should fix it? I think. Unless I've messed up my reasoning (quite possible) or something funky is going on with your setup (hopefully not).

Would you be willing to try out the fix I mentioned earlier, https://github.com/dlukes/ultisnips/commit/e3ff99d2d28f07a10dddae5a292414ba771e4a0d? Does it get rid of the duplicates for you?

PR #1447, opened 2022-03-04
Pass through trigger key if no expansion done

### LINKED ISSUE #1184: UltiSnips should pass through any mapping that it currently can't execute as the trigger key
Ultisnips plays very good with <Tab> for selection but when I map <CR> for selection, <CR> no longer triggers newline in insert mode. It would be nice to have <CR> do both as <tab> does because it feels natural to step through the completion suggestions (YCM here) using <tab> and ultisnips stops at the first snip suggestion and starts the snip while using <tab>

1. `let g:UltiSnipsExpandTrigger = "<CR>"`
2. in insert mode, some text `<CR>`
3. no newline

PS: There seems to be one possible solution on stackoverflow: https://vi.stackexchange.com/questions/2757/triggering-ultisnips-with-enter but it would be nice to have that in-built.
-----

- **Operating System**:  Archlinux
- **Vim Version**: VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 26 2020 11:45:38)    
Included patches: 1-148 
- **UltiSnips Version**: 129d33fc34a72ced857f3cfb72ff10caf10f0020 
- **Python inside Vim**: 3.8.1

richyliu:
Fixes #1184.

- if the trigger key does not expand a snippet, then pass it through
  like a normal key
- if the expand/jump key use the same trigger, then the key will be
  passed through if there is no expansion or jump
- added test cases

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -177,7 +177,7 @@ class SnippetManager:
         vim_helper.command("let g:ulti_expand_res = 1")
         if not self._try_expand():
             vim_helper.command("let g:ulti_expand_res = 0")
-            self._handle_failure(self.expand_trigger)
+            self._handle_failure(self.expand_trigger, True)
 
     @err_to_scratch_buffer.wrap
     def expand_or_jump(self):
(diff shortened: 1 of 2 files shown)


SirVer:
Thanks! @richyliu Do you have any why the github workflow is not running on this branch? I'd like to see the test suite on this though.

richyliu:
I pushed another commit which triggered it.

richyliu:
It seems likes "Build & Test using Docker (8.1, 3.8-buster, vim_81_py38)" is getting stuck, not sure why.

SirVer:
@richyliu I do not know why either, but it is the same in master. I have spend some time on in, but have not figured it out up until now. Thanks for the contribution, gonna merge!

PR #1457, opened 2022-04-24
Prettify README

EmilySeville7cfg:


.github/workflows/main.yml
```diff
@@ -1,3 +1,4 @@
+name: CI
```
SirVer, line 1: Can you explain what this does?

.replit
```diff
@@ -0,0 +1,2 @@
+language = "python3"
```
SirVer, line 1: I have no idea what his file does.

README.md
```diff
@@ -1,19 +1,38 @@
 ![Build Status](https://github.com/SirVer/ultisnips/actions/workflows/main.yml/badge.svg)
```
SirVer, line 1: These are very opionated changes. Can you outline your thought process around each one?

EmilySeville7cfg: I wanted to prettify README and structure it better. My goal was to follow this template:

- badges: CI status, official chat, doc reference
- short project description and maybe some important notes (for instance where to get snippets in this case)
- main features
- integrations
- usage example
- installation (via Vim package system or custom package manager)
- defaults
- tutorials
- history notes (for those users who are really interested in them)

SirVer: Thanks for your suggestion, however I do not agree with this structure being better than the current README. I really appreciate you taking the time!

PR #1462, opened 2022-05-18
Correct snipGlobal syntax regions

wookayin:
The previous syntax group for snipGlobal was not matching and including
the endglobal keyword, resulting in a wrong folding region.

This commit fixes this, to make the endglobal keyword matched and
correctly included in the snipGlobal syntax region.

### Wrong (before fix):

image

Note L18 `endglobal` is excluded and therefore not folded (with `foldmethod=syntax`).

### Fixed (after):

image

### DIFF
--- a/syntax/snippets.vim
+++ b/syntax/snippets.vim
@@ -133,16 +133,16 @@ syn cluster snipTabStopTokens add=snipTransformation
 
 " Generic (non-Python) {{{4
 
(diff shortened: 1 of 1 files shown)


wookayin:
Similar, related, but a bit different: #1463

SirVer:
Thank you!

PR #1463, opened 2022-05-18
Fix snipSnippetContextP syntax to match the ending double quote

wookayin:
Excluding the double quote from match in `snipSnippetContextP` groups
can result in a wrong syntax highlight.

Therefore the ending `"` should be included in the syntax group.

Before:

image

After:

image


SirVer:
Thanks!

Issue #1514, opened 2023-01-11
vim_interface's has_version() doesn't handle non-contiguous patch ranges properly


jamessan:
https://github.com/SirVer/ultisnips/blob/e99fdf15cd55a4a8e0cb0a80a6810c1867a5c401/test/vim_interface.py#L92-L97

This errors out with a version like:

```
$ vim --version | grep Included
Included patches: 1-1000, 1087, 1117-1118, 1129
```

```
     _patch = int(line.split(":")[-1].strip().split("-")[-1])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '1118, 1129'
```

The simplistic version checking here also implies the wrong semantics.  A patch is only included if `(_major, _minor) > (major, minor)` or if `patch` is one of the listed patch levels.
Also, checking if the highest patch number is greater than the given patch isn't a valid way to see if a patch is included.  In this example, `has_version(9, 0, 1088)` should return False.

SirVer:
@jamessan Thanks for reporting this. Are you interested in crafting a patch?

jamessan:
I'm not familiar with the code, so no.  I was just forwarding the issue from Debian.

PR #1516, opened 2023-01-13
Check Vim version through call to Vim's has() function

mymedia2:
It looks Vim 9.0 changed format output of `--version` and no longer prints applied extra patches on a separate line but joins the patch numbers with comma. As @jamessan rightly noted, the has_version method should check whether `patchlevel` is in set of listed patches.

And it is worth mentioning that the exact format is not documented and may change in future.

Fixes: #1514

### DIFF
--- a/test/vim_interface.py
+++ b/test/vim_interface.py
@@ -76,27 +76,18 @@ class VimInterface(TempFileManager):
     def __init__(self, vim_executable, name):
         TempFileManager.__init__(self, name)
         self._vim_executable = vim_executable
-        self._version = None
 
     @property
     def vim_executable(self):
         return self._vim_executable
 
     def has_version(self, major, minor, patchlevel):
-        if self._version is None:
-            output = subprocess.check_output([self._vim_executable, "--version"])
-
-            _major = 0
-            _minor = 0
-            _patch = 0
-            for line in output.decode("utf-8").split("\n"):
-                if line.startswith("VIM - Vi IMproved"):
-                    _major, _minor = map(int, line.split()[4].split("."))
-                if line.startswith("Included patches:"):
-                    _patch = int(line.split(":")[-1].strip().split("-")[-1])
-            self._version = (_major, _minor, _patch)
-
-        return self._version >= (major, minor, patchlevel)
+        cmd = [
+            self._vim_executable, "-e", "-c",
+            "if has('patch-%d.%d.%d') | quit | else | cquit | endif"
+                % (major, minor, patchlevel),
+        ]
+        return not subprocess.call(cmd, stdout=subprocess.DEVNULL)
 
     def get_buffer_data(self):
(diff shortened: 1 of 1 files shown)


test/vim_interface.py
```diff
@@ -85,18 +86,25 @@ def vim_executable(self):
...
-        return self._version >= (major, minor, patchlevel)
+                    for chain in line.split(":")[-1].split(","):
+                        if "-" in chain:
+                            lower, higher = map(int, chain.split("-"))
+                            self._patches.update(range(lower, higher))
+                        else:
+                            self._patches.add(int(chain))
+                if line.startswith("Extra patches:"):
+                    # Vim before 9.0 also prints extra pacthes as full version
+                    # numbers. Example: 8.2.3402, 8.2.3403, 8.2.3409, 8.2.3428
```
jamessan, line 101: This is not a change in behavior.  There are two different ways that Vim tracks patches.

1. [Included patches](https://github.com/vim/vim/blob/b40a2fb3b3e4fef137eebb3bab9c02adc5540658/src/version.c#L696) are the different patch levels for the current version that are included.
2. [Extra patches](https://github.com/vim/vim/blob/b40a2fb3b3e4fef137eebb3bab9c02adc5540658/src/version.c#L3090-L3097) are free-form strings that aren't usually used in official Vim builds, but can be used in other builds.  For Debian, that's where I document patches that are back ported from a different release (e.g, applying 8.2.3402 to an 8.1 build of Vim).

Since there's no defined structure for extra patches, I'm not sure it's worth parsing.

test/vim_interface.py
```diff
@@ -85,18 +86,25 @@ def vim_executable(self):
     def has_version(self, major, minor, patchlevel):
         if self._version is None:
             output = subprocess.check_output([self._vim_executable, "--version"])
```
jamessan, line 88: What about using Vim's own logic to detect if a patch is included?  You could do something like

```python
output = subprocess.check_output([self._vim_executable, '--cmd', f'echo has("patch-{major}.{minor}.{patchlevel}")', '--cmd', 'quit'])
```

assuming the minimum supported vim version is >= 7.4.237.

Or if it's easier to rely on the exit code

```python
has_version = subprocess.call([self._vim_executable, '--cmd', f'if has("patch-{major}.{minor}.{patchlevel}") | quit | else | cquit | endif'])
```

`:help has-patch` describes an alternative that works for older Vim versions.

mymedia2:
@jamessan Thank you for the review and the explanation. I rewrote the patch to call of the Vim's original has() function and the has_version() method in Python code became shorter.

jamessan (approved):

SirVer:
Wonderful! Thanks for the contribution.

SirVer:
Also thanks to @jamessan for reviewing. What an exemplary piece of open source interaction this contribution was - only I stick out like a sore thumb coming super late to press the merge button. Thanks you both!

PR #1541, opened 2023-09-25
Mark failing CI tests as failed

yut23:
With `pipefail`, the exit code for a pipeline will be taken from the first failing command in the pipeline, rather than the last command.


SirVer:
Thank you very much!

PR #1542, opened 2023-09-25
Add JumpOrExpand support

pilgrimlyieu:
This is a very small improvement. As far as I know, we can not consider to jump forward instead of expanding snippets first if the expand trigger is as same as the jump forward trigger. So I introduce two new options `g:UltiSnipsExpandOrJumpTrigger` & `g:UltiSnipsJumpOrExpandTrigger` to achieve this goal. If none of them is set, it'll still expand snippets first if two triggers remain same, or it'll map separately(that is, same as before, nothing broken).

Besides, I'm wondering how to write the document. I'll appreciate if some one could help me improve the document with adding the introduction of these two new options.

I'm not familiar with this project that I may misunderstand something or make some mistakes. If so, please let me know! Any advice is welcome. Hopefully my PR could be accepted.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -284,7 +284,7 @@ is set, g:UltiSnipsExpandOrJumpTrigger variable will take effects.
 
 3.2.2 Using your own trigger functions           *UltiSnips-trigger-functions*
 
-For advanced users there are four functions that you can map directly to a
+For advanced users there are functions that you can map directly to a
 key and that correspond to some of the triggers previously defined:
    g:UltiSnipsExpandTrigger        <--> UltiSnips#ExpandSnippet
    g:UltiSnipsJumpForwardTrigger   <--> UltiSnips#JumpForwards
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -246,6 +246,7 @@ The priority will then be rails -> ruby -> programming -> all.
 
                            *g:UltiSnipsExpandTrigger* *g:UltiSnipsListSnippets*
                *g:UltiSnipsJumpForwardTrigger* *g:UltiSnipsJumpBackwardTrigger*
+              *g:UltiSnipsExpandOrJumpTrigger* *g:UltiSnipsJumpOrExpandTrigger*
 
 You can define the keys used to trigger UltiSnips actions by setting global
 variables. Variables define the keys used to expand a snippet, jump forward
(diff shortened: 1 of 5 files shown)


SirVer:
Thanks for your contribution! 

You understood correctly that the functionality is not there and I think your approach for implementing it is reasonable. This needs testing and update to the documentation.

As for the documentation, you just change the doc/UltiSnips.txt - it is all manual. For testing see the Contributing document.

pilgrimlyieu:
I'm not a English native speaker so my expression may be unclear and vague(in this PR & in the documentation).

In addition, I'm sorry that I don't know how to add a test because `EX` standing for expanding snippets and `JF` standing for jumping forward in `constant.py` have been distinguished. Maybe I've missed something? Could you give me more specific details?

doc/UltiSnips.txt
```diff
@@ -267,27 +268,45 @@ following to your vimrc file or switching to a plugin like Supertab or
...
+Or else, if you have a preference for jumping forward first, set
+g:UltiSnipsJumpOrExpandTrigger instead. >
+   let g:UltiSnipsJumpOrExpandTrigger = "<tab>"
+
+If both g:UltiSnipsExpandOrJumpTrigger and g:UltiSnipsJumpOrExpandTrigger
+is set, g:UltiSnipsExpandOrJumpTrigger variable will take effects.
+
 3.2.2 Using your own trigger functions           *UltiSnips-trigger-functions*
 
 For advanced users there are four functions that you can map directly to a
```
SirVer, line 287: no longer four - please just change to "there are functions" which is more future proof

pythonx/UltiSnips/snippet_manager.py
```diff
@@ -197,6 +197,24 @@ def expand_or_jump(self):
             vim_helper.command("let g:ulti_expand_or_jump_res = 0")
             self._handle_failure(self.expand_trigger, True)
 
+    @err_to_scratch_buffer.wrap
+    def jump_or_expand(self):
+        """This function is used for people who wants to have the same trigger
```
SirVer, line 202: nit: people who want

SirVer:
I am not a native speaker either, but your english seems excellent from all that I can tell. The documentation is very well like that, but this still needs a test, otherwise it likely will regress the functionality in the next refactor.

I think you want to add a test in [this file](https://github.com/SirVer/ultisnips/blob/master/test/test_Expand.py#L71). The idea is that you use  `_extra_vim_config` to set one of your new variables, then define two snippets, one that you expand, then another one that creates ambiguity between jumping and expanding, then you check that the right thing happened.

doc/UltiSnips.txt
pilgrimlyieu: Fixed.

pythonx/UltiSnips/snippet_manager.py
pilgrimlyieu: Fixed. It's "people who wants" in `expand_or_jump` function and I fix it too.

pilgrimlyieu:
I've run the tmux test in WSL and it seems that tests I added have passed. Did I do correctly? What else should I do?

SirVer:
Absolutely great contribution. Thank you very much for this!

Konfekt:
Thank you very much! I was ever since missing this feature that e.g. neosnippet had and had to work around it by something like (using `Ctrl-E` as a trigger)

```vim
let g:ulti_expand_or_jump_res = 0
function! UltiSnips_ExpandableOrJumpable()
  call UltiSnips#ExpandSnippetOrJump()
  if g:ulti_expand_or_jump_res > 0
    return ""
  else
    if pumvisible()
      return "\<C-E>"
    else
      return "\<End>"
    endif
  endif
endfunction

inoremap <silent> <C-E> <c-r>=UltiSnips_ExpandableOrJumpable()<cr>
```

Konfekt:
From the commit itself it is not obvious to me what happens in case neither a jump nor expansion is possible. Does the mapping fall back to the default operation of the supplied key combo or to no operation?

pilgrimlyieu:
> From the commit itself it is not obvious to me what happens in case neither a jump nor expansion is possible. Does the mapping fall back to the default operation of the supplied key combo or to no operation?

The first case, falling back to the default operation. For example if I set

```vim
let g:UltiSnipsJumpOrExpandTrigger = "<Tab>"
```

It'll jump or exapnd if it could, or it will send <kbd>Tab</kbd>.

PR #1551, opened 2024-01-26
Add variable to control AutoTrigger

jdujava:
~Now it is possible to disable AutoTrigger in the current buffer by setting the buffer variable `b:UltiSnipsAutoTrigger = 0`.~

Now it is possible to disable AutoTrigger via global variable `g:UltiSnipsAutoTrigger`, and to dynamically toggle AutoTrigger on/off with `Ultisnips#ToggleAutoTrigger()`.

This is just a first crude shot at implementing this "toggling feature" (documentation and so on TODO).
What do you think, is this alright, or is there perhaps some better way in your opinion?

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -123,6 +123,11 @@ function! UltiSnips#CanJumpBackwards() abort
 	return can_jump_backwards
 endfunction
 
+function! UltiSnips#ToggleAutoTrigger() abort
+    py3 vim.command("let autotrigger = %d" % UltiSnips_Manager._toggle_autotrigger())
+    return autotrigger
+endfunction
+
 function! UltiSnips#SaveLastVisualSelection() range abort
     py3 UltiSnips_Manager._save_last_visual_selection()
     return ""
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -511,6 +511,21 @@ situation. This is useful in conditional mappings.
 This function returns 1 if UltiSnips can jump backwards in the current
 situation. This is useful in conditional mappings.
 
+ 3.4.7 UltiSnips#ToggleAutoTrigger              *UltiSnips#ToggleAutoTrigger*
+
+This function toggles the `autotrigger` functionality. Its return value
+corresponds to the new state of the `autotrigger` (0: disabled, 1: enabled).
+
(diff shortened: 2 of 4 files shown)


SirVer:
I worry a bit about performance. Calling out to vim.eval is slow and this code will run on every cursor movement. This should probably be checked in vimscript before calling into python. Overall, I am fine with the feature.

ces42:
> I worry a bit about performance. Calling out to vim.eval is slow and this code will run on every cursor movement. This should probably be checked in vimscript before calling into python. Overall, I am fine with the feature.

Yes this can be done much better by removing/re-creating the `UltiSnips_AutoTrigger` autogroup from `plugin/UltiSnips.vim` using vimscript. There's no need to write python code for this.

jdujava:
To not complicate things, I added only a global toggling mechanism.

@ces42 To be honest, I wasn't sure wheter `UltiSnips_Manager._track_change()` was needed also for other purposes than AutoTrigger, so I tried it this way. But I am certainly open to other (possibly better) solutions!

ces42:
Adding this to `autoload/UltiSnips.vim` provides control over autotrigger:
```vim
function! UltiSnips#EnableAutotrigger() abort
    augroup UltiSnips_AutoTrigger
        au!
        au InsertCharPre * call UltiSnips#TrackChange()
        au TextChangedI * call UltiSnips#TrackChange()
        if exists('##TextChangedP')
            au TextChangedP * call UltiSnips#TrackChange()
        endif
    augroup END
endfunction

function! UltiSnips#DisableAutotrigger() abort
    augroup UltiSnips_AutoTrigger
        au!
    augroup END
endfunction

function! UltiSnips#ToggleAutotrigger() abort
    if exists('#UltiSnips_AutoTrigger')
        call UltiSnips#DisableAutotrigger()
    else
        call UltiSnips#EnabgleAutotrigger()
    endif
endfunction
```

SirVer:
Great, but a test is still missing. Will you work on that @jdujava ?

jdujava:
@SirVer Added tests and documentation, does it look OK?

jdujava:
@SirVer is it good to merge, or is there perhaps anything else to do?

SirVer (approved):

SirVer:
@jdujava Yes, it looks great, an exemplary contribution to an open source project. Thank you very much! Sorry that I took so long to get around to it.

Issue #1496, opened 2022-10-13
ultisnips causing slowness displaying text in insert mode


rafaelleru:
neovim has been displaying text while writting for a couple of days, after upgrading it to the latest version and trying multiple versions I tried to profile neovim to see what happened. Ultisnips function UltiSnips#TrackChange() is taking near 3 seconds to run, causing a big delay in text rendering in insert mode.

**Expected behavior:**
vim to be as fast as always

**Actual behavior:**
vim is too slow to render text in insert mode

**Steps to reproduce**
load teh extension
write text on any file

-----

- **Operating System**:  Ubuntu 22.04
- **Vim Version**: neovim 0.8.0
- **UltiSnips Version**: latest
- **Python inside Vim**:  3.10.6
- **Docker repo/vimrc**:  https://github.com/rafaelleru/dotfiles/tree/ansible/config/nvim

# Output of the profile file for the given function:

```
FUNCTION  UltiSnips#TrackChange()
    Defined: ~/.vim/plugged/ultisnips/autoload/UltiSnips.vim:176
Called 56 times
Total time:   2.833268
 Self time:   2.833268

count  total (s)   self (s)
   56              2.833192     py3 UltiSnips_Manager._track_change()

FUNCTION  provider#python3#Call()
    Defined: /usr/local/share/nvim/runtime/autoload/provider/python3.vim:26
Called 56 times
Total time:   2.832244
 Self time:   2.832244

count  total (s)   self (s)
   56              0.000319   if s:err != ''
                                return
   56              0.000070   endif
   56              0.000257   if !exists('s:host')
                                let s:rpcrequest = function('rpcrequest')
                            
                                " Ensure that we can load the Python3 host before bootstrapping
                                try
                                  let s:host = remote#host#Require('legacy-python3-provider')
                                catch
                                  let s:err = v:exception
                                  echohl WarningMsg
                                  echomsg v:exception
                                  echohl None
                                  return
                                endtry
   56              0.000029   endif
   56              2.830506   return call(s:rpcrequest, insert(insert(a:args, 'python_'.a:method), s:host))


```

* H

kuator:
neovim is not supported

SirVer:
Fixed in #1552

PR #1552, opened 2024-03-06
fix #1165: insert mode lag

### LINKED ISSUE #1165: Insert mode lags
In insert mode I can see quite a lag especially when I do enter Russian letters. 

**Expected behavior:**
Should be no lag

**Actual behavior:**
Lag

ultisnips-issue1

**Steps to reproduce**

Well, not sure. I just had it installed for quite a long time

Settings that I had regarding ultisnips
```
	let g:UltiSnipsExpandTrigger = '<tab>'
	let g:UltiSnipsJumpForwardTrigger = '<tab>'
	let g:UltiSnipsJumpBackwardTrigger = '<s-tab>'
	let g:UltiSnipsListSnippets = '<c-tab>'
```

My snippets were:

https://github.com/habamax/.vim/tree/03dce7b57131e993b5ccdc88775196c2f5fe7ef6/UltiSnips

If I remove Ultisnips (use vsnips instead):
ultisnips-issue3

PS

If I remove `au! Ultisnips_AutoTrigger` it becomes way faster. But only for english. Russian lags the same.

ultisnips-issue2

-----

- **Operating System**: Win10 
- **Vim Version**: 8.2.114 
- **UltiSnips Version**: current master
- **Python inside Vim**: 3.8.1

mg979:
Ultisnips was traversing the whole runtimepath every time an expansion was attempted, something that happens for every key press since there is autotrigger.

Even without autotrigger, traversal was done at every snippet expansion, which caused slowdowns in that case.

With this change, traversal happens only the first time, or after a source must be refreshed (because a snippets file has been updated).

Note: I cannot run tests (Windows).

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -3,6 +3,9 @@ if exists("b:did_autoload_ultisnips")
 endif
 let b:did_autoload_ultisnips = 1
 
+" Ensure snippets are loaded for current buffer
+au UltiSnips_AutoTrigger FileType,BufEnter * call UltiSnips#CheckFiletype()
+
 " Also import vim as we expect it to be imported in many places.
 py3 import vim
 py3 from UltiSnips import UltiSnips_Manager
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -183,6 +183,10 @@ function! UltiSnips#TrackChange() abort
     py3 UltiSnips_Manager._track_change()
 endfunction
 
+function! UltiSnips#CheckFiletype() abort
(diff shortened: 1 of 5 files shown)


mg979:
The test `test_SnipMate.snipMate_OtherFiletype` fails, I don't know why. If I do the steps manually in the editor it works. Any idea?

mg979:
Should be fixed now, it was broken because it only worked for the first loaded filetype.

SirVer:
@mg979 Thanks for working on this, Gianmaria! This was a common complaint and I never found the time to track it down. Your work is very appreciated and thanks for fixing the tests!

mg979:
The initial commit broke proper snippets loading actually, not just tests. I think #1496 could also be closed.

PR #1616, opened 2026-04-18
test: bump nvim sleeptime to avoid mode-glitch races

SirVer:
Neovim's Python plugin is a remote process communicating via msgpack RPC, so every CursorMovedI handler adds ms-level latency on top of whatever UltiSnips does in Python. At the default sleeptime=0 the integration test runner's tmux keystrokes race ahead of the handler and nvim ends up with mode glitches — most visibly, the ESC in
`get_buffer_data()`'s "ESC + :w! <path>\n" doesn't take effect, so the write command gets typed into the buffer instead of executed.

The test framework retries with progressively higher sleeptimes until a run succeeds, which is why the nvim suite takes much longer than the Vim one.

Give the nvim path more headroom on the first attempt (0.05 instead of 0.00) and start the retry ladder at 0.15 instead of 0.01. The Vim path is unchanged.

Results on my machine:

nvim full suite: 1517s → 793s  (−48%)
vim  full suite: 145s (baseline, unchanged)

Gap to Vim narrowed from ~10x to ~5.5x.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,23 @@ MAKEFILE_DIR := $(dir ${MAKEFILE_PATH})
 image_repro:
 	docker build -t ultisnips:repro --build-arg PYTHON_IMAGE=3.13-bookworm --build-arg VIM_VERSION=git .
 
+# Neovim variant.  Uses Dockerfile.nvim (same one CI uses for nvim jobs)
+# with the version pinned to what CI tests.  Run `make image_repro_nvim`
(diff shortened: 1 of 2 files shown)


SirVer:
This was not really consistent - on CI it was much slower. I was experimenting feeding data directly into pynvim - which was faster, but felt to me to deviate too far from an actual user typing keys in, so I abandoned this work and decided to live with the status quo.

PR #1618, opened 2026-04-24
Document that jump trigger mappings are only live while a snippet is active

### LINKED ISSUE #1497: Verbose map of <C-j> and <C-k> according to documented settings does not work as expected
**Expected behavior:**

I have the following vimrc:

```
if empty(glob('~/.vim/autoload/plug.vim'))
    silent !curl -fLo ~/.vim/autoload/plug.vim --create-dirs
        \ https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
    autocmd VimEnter * PlugInstall --sync | source $MYVIMRC
endif
call plug#begin('~/.vim/plugged')
Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
call plug#end()

let g:UltiSnipsExpandTrigger = '<C-j>'
let g:UltiSnipsJumpForwardTrigger = '<C-j>'
let g:UltiSnipsJumpBackwardTrigger = '<C-k>'

```

This is similar to what is specified in the official documentation here: https://github.com/SirVer/ultisnips/blob/master/doc/UltiSnips.txt

Once I load this vimrc and open a .tex file, I issue the following commands:

`:verbose map <C-j>`

I expect this to specify that <C-j> is mapped to trigger expansion and trigger forward jump.

Likewise, I expect `:verbose map <C-k>` to indicate that <C-k> is mapped to trigger backward jump.

**Actual behavior:**
While `:verbose map <C-j>` provides expected output that indicates that it has read in the vimrc, `:verbose map <C-k>` returns that no mapping has been found.

**Steps to reproduce**

1. The entire vimrc and sequence of steps has been provided above.
2. In addition, here is an animated gif that shows the steps I followed and it should be easy to reproduce: https://imgur.com/qkHau09
-----

- **Operating System**: WSL Ubuntu 20.04 on Windows
- **Vim Version**: VIM - Vi IMproved 8.1 (2018 May 18, compiled Apr 15 2020 06:40:31)
Included patches: 1-2269
- **UltiSnips Version**: 53e1921e3ef015ef658e540c0aa9c4835f9c18a6
- **Python inside Vim**: 9,78           All3.8.10 (default, Jun  2 2021, 10:49:15) [GCC 9.4.0]
- **Docker repo/vimrc**: vimrc provided above

SirVer:
Extend the existing note in `doc/UltiSnips.txt` so it explicitly covers
the common `:verbose map <c-k>` confusion reported in #1497: the jump
triggers are mapped lazily as buffer-local mappings only while a snippet
is currently being expanded, so inspecting them with `:map` reports
nothing until a snippet is active.

No code change — current behavior is intentional (documented since the
note was introduced) so that UltiSnips doesn't clobber the user's own
`<c-j>` / `<c-k>` mappings outside of snippet expansion.

Fixes #1497.

### DIFF
--- a/doc/UltiSnips.txt
+++ b/doc/UltiSnips.txt
@@ -263,7 +263,10 @@ their default values are: >
    g:UltiSnipsJumpBackwardTrigger         <c-k>
 
 UltiSnips will only map the jump triggers while a snippet is active to
-interfere as little as possible with other mappings.
+interfere as little as possible with other mappings. Because of this,
+inspecting the mappings with |:map| (for example `:verbose imap <c-k>`)
+reports no mapping until a snippet is actually being expanded — expand
+a snippet first, then inspect the mappings while still inside it.
 
 The default value for g:UltiSnipsJumpBackwardTrigger interferes with the
 built-in complete function: |i_CTRL-X_CTRL-K|. A workaround is to add the


SirVer:
Actually, I now think this doc update is silly.

Issue #1524, opened 2023-02-27
VISUAL in post_jump  - can we access? How


JoseConseco:
I see no discussion section on ultisnips page, so I post my question here:
how do I access VISUAL  (or vim yank register) from **post_jump** event? 

I want snippet for extracting variable e.g. - `a+b`:
`  something(a + b)`
selecting content of bracket and running snipped would generate:
```
new_var = a+b
something(new_war)
```

I tried to access snip.v.text or snip.visual_text in  post_jump : 
`post_jump "if snip.tabstop == 0: insert_method_call(snip.tabstops[1].current_text, snip.visual_text)" `

But ultisnips complains about snip not having   `v`   or `visual_text`  props.

**My Code**

```
global !p
def post(var_name, visual):
	snip.buffer[snip.snippet_start[0]:snip.snippet_start[0]] = [var_name + ' = ' + visual] 
endglobal
post_jump "if snip.tabstop == 0: post(snip.tabstops[1].current_text, snip.visual_text)"
snippet var "extract var" b
$1
endsnippet 
```
I also tried to use ${VISUAL} in expand_anon,   but no luck.

Finally there is vim  - getreg() :
vim.command(":call append(line('.')-1,   '"+ var_name + " = '.getreg('+'))")
But then Ultisnip complains about modifying buffer using vim.command.

SirVer:
This is currently not supported I think, a patch + test case would be very welcome

paulfioravanti:
My workaround for this is to store the `snip.visual_text` in the `context`, to then be accessed later during `post_jump`. Here is an example:

```snippet
global !p

def store_visual_text(snip):
    return {"visual": snip.visual_text}

def use_visual_text(snip):
    visual = snip.context["visual"].strip(" \t\n\r")
    snip.expand_anon("Visual was: ${1:" + visual + "}")

endglobal

context "store_visual_text(snip)"
post_jump "use_visual_text(snip)"
snippet visual "test post_jump visual output"
$0
endsnippet
```

[...]

PR #1632, opened 2026-05-10
Fix: dark-factory/fix-github-SirVer-ultisnips-1524

andreizahorski1-code:
Automated fix from andreizahorski1-code/ultisnips:dark-factory/fix-github-SirVer-ultisnips-1524

### DIFF
--- a/pythonx/UltiSnips/snippet/definition/base.py
+++ b/pythonx/UltiSnips/snippet/definition/base.py
@@ -461,6 +461,11 @@ class SnippetDefinition:
                 "tabstops": current_snippet.get_tabstops(),
                 "snippet_start": start,
                 "snippet_end": end,
+                # Passed visual_content and current_placeholder from the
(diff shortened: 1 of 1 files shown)


SirVer:
Lacking tests.

SirVer:

The direction matches #1524 (which I'd asked for a patch on), but the change needs more work before it can land:

- **No test.** The project requires an integration test for new behavior — see [`docs/CONTRIBUTING.md`](https://github.com/SirVer/ultisnips/blob/master/docs/CONTRIBUTING.md).
- `current_snippet.visual_content` is `None` when the snippet wasn't expanded over a visual selection; `visual_content.mode` will raise `AttributeError` in that path.
- The keys are exposed only in `do_post_jump`. The same data is just as relevant for `do_pre_expand` / `do_post_expand` — an asymmetric API surface is worse than none.

Please open a new PR with tests and a complete API surface if you'd like another pass.

Issue #1519, opened 2023-02-04
 Multiple match dialog shown for same snippet


Silur:
Opening a new buffer `/tmp/test.jsx` (or any other file type) and trying a snippet from vim-snippets shows the "multiple matches" selection dialog, even though it shows that both options are in fact coming from the same file, from the same line:

```
# vim /tmp/test.jsx

rfc<tab>

1: (rfc) "react functional component" (/home/silur/.vim/pack/vendor/start/vim-snippets/UltiSnips/javascript_react.snippets:10)
2: (rfc) "react functional component" (/home/silur/.vim/pack/vendor/start/vim-snippets/UltiSnips/javascript_react.snippets:10)
Type number and <Enter> or click with the mouse (q or empty cancels): 

```

Selecting either option will complete the snippet as usual.

Asciinema demonstration: https://asciinema.org/a/6lyePLGN28jhYlI1f5kzqlXJo

```
VIM - Vi IMproved 9.0 (2022 Jun 28, compiled Jan 12 2023 11:41:18)
Included patches: 1-1182
Compiled by Arch Linux
Huge version without GUI.  Features included (+) or not (-):
+acl               +file_in_path      +mouse_urxvt       -tag_any_white
+arabic            +find_in_path      +mouse_xterm       +tcl/dyn
+autocmd           +float             +multi_byte        +termguicolors
+autochdir         +folding           +multi_lang        +terminal
-autoservername    -footer            -mzscheme          +terminfo
-balloon_eval      +fork()            +netbeans_intg     +termresponse
+balloon_eval_term +gettext           +num64             +textobjects
-browse            -hangul_input      +packages          +textprop
++builtin_terms    +iconv             +path_extra        +timers
+byte_offset       +insert_expand     +perl/dyn          +title
+channel           +ipv6              +persistent_undo   -toolbar
+cindent           +job               +popupwin          +user_commands
-clientserver      +jumplist          +postscript        +vartabs
-clipboard         +keymap            +printer           +vertsplit
+cmdline_compl     +lambda            +profile           +vim9script
+cmdline_hist      +langmap           -python            +viminfo
+cmdline_info      +libcall           +python3/dyn       +virtualedit
+comments          +linebreak         +quickfix          +visual
+conceal           +lispindent        +reltime           +visualextra
+cryptv            +listcmds          +rightleft         +vreplace
+cscope            +localmap          +ruby/dyn          +wildignore
[...]

PR #1633, opened 2026-05-10
Fix: dark-factory/fix-github-SirVer-ultisnips-1519

andreizahorski1-code:
Automated fix from andreizahorski1-code/ultisnips:dark-factory/fix-github-SirVer-ultisnips-1519

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -62,6 +62,20 @@ def _show_user_warning(msg):
 def _ask_snippets(snippets):
     """Given a list of snippets, ask the user which one they want to use, and
     return it."""
+    # Deduplicate entries that are identical in display attributes.
+    # This works around cases where internal filtering might miss subtle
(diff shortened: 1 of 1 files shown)


SirVer:
@andreizahorski1-code 

Please stop these low quality spam contributions. I appreciate fixes, but they need to have a proper MR description, come with a test case that fails before and passes after and show in general some understanding of what is happening. This is not helping.

SirVer:

#1519 is about the same snippet file being loaded twice from `&runtimepath` — note the issuer's example shows both entries pointing at the **same `(file, line)`**. The bug is in snippet *loading*, not display. Deduping inside `_ask_snippets` hides the underlying load-time bug, and it also risks suppressing legitimate disambiguation cases where two distinct snippets happen to share `(trigger, description, location)`.

The patch's own comment — "works around cases where internal filtering might miss subtle differences" — concedes that this is a workaround, not a fix.

There's also no test.

The right place to fix this is in snippet loading. Closing.

Issue #1508, opened 2022-11-23
when snip list appear, the first snip is auto selected. if I input "enter", the first snip is not being expanded


aizhiqi:
**Expected behavior:**
let g:UltiSnipsExpandTrigger = "<cr>"

when snip list appear, the first snip is auto selected. if i input "enter", the first snip is expanded

**Actual behavior:**
let g:UltiSnipsExpandTrigger = "<cr>"

when snip list appear, the first snip is auto selected. if I input "enter", the first snip is not being expanded

**Steps to reproduce**
1.insert follow line to config:
let g:UltiSnipsExpandTrigger = "<cr>"

2.vim test.py

3.input:"de"

4.vim auto triger snip list and the first line is seleced

5.input "enter", but the first snip is not being expanded

1.
2.
3.

-----

- **Operating System**: 
- **Vim Version**: 
- **UltiSnips Version**: 
- **Python inside Vim**: 
- **Docker repo/vimrc**:

SirVer:
This is two separate things, neither of which is an UltiSnips bug:

1. **"The first item is auto-selected in the popup."** That popup isn't UltiSnips — UltiSnips doesn't open a completion popup as you type. It's coming from whatever completion plugin you have (coc.nvim / nvim-cmp / YouCompleteMe / deoplete / …) which pulls UltiSnips in as a source. Whether the first entry is preselected is controlled by Vim's `completeopt`. If you don't want preselection, add `noselect`:

   ```vim
   set completeopt=menu,menuone,noselect
   ```

2. **"`<CR>` doesn't expand the highlighted snippet."** `g:UltiSnipsExpandTrigger = "<cr>"` makes Enter call `UltiSnips#ExpandSnippet()`, which only looks at the text immediately before the cursor (here: `de`) and tries to match a snippet *trigger*. It has no idea which item your completion plugin highlighted in the popup, and popup entries are typically shown by description, not trigger — so Enter quite reasonably does nothing.

   Getting `<CR>` to mean "accept popup item → then expand if it's a snippet" is the well-known three-way collision between Enter, popup-accept, and snippet-expand. The fix lives in your completion plugin's docs, not in UltiSnips:
   - coc.nvim: `coc#pum#confirm()` / the `<CR>` recipe in `:h coc-completion`
   - nvim-cmp: the `cmp.mapping.confirm({ select = true })` example in the README
   - YouCompleteMe / others: their respective integration sections

[...]

PR #1634, opened 2026-05-10
Fix: dark-factory/fix-github-SirVer-ultisnips-1508

andreizahorski1-code:
Automated fix from andreizahorski1-code/ultisnips:dark-factory/fix-github-SirVer-ultisnips-1508

### DIFF
--- a/pythonx/UltiSnips/snippet_manager.py
+++ b/pythonx/UltiSnips/snippet_manager.py
@@ -285,6 +285,12 @@ class SnippetManager:
         if not snippets:
             return True
 
+        # If only one snippet is found, expand it immediately without asking,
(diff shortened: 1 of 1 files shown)


SirVer:
Lacking test.

SirVer:

This change doesn't fix #1508. #1508 is about the **completion popup** auto-selecting an entry and `<cr>` being consumed by the popup before UltiSnips sees it — a completely different code path. `list_snippets` is not involved in that flow.

The patch also has an unintended side effect: it changes the semantics of `:UltiSnipsListSnippets` for users who call it deliberately to browse. They would now get a silent expansion whenever there is exactly one match, instead of seeing the listing.

There's also no test.

Closing — please re-read the issue.

PR #1638, opened 2026-05-12
Fix junk characters appearing in snippet expansions

### LINKED ISSUE #751: Junk characters in snippet expansion text
I have a problem with a somewhat evil snippet which looks like this:

```python
snippet "rlink(?:\s+\#([-\w]+)?)?\s+(\S+)(?:\s+(.*))?" "save an external reflink" r
[${1:`!p snip.rv = reg_or(snip.v.text, reg_or(match.group(3), 'LINK_TEXT'))`}][`!p
id = reg_or(match.group(1), t[1], filtr=html_id)
url = reg_or(match.group(2), '!!!URL-MISSING!!!')
title = reg_or(match.group(3), "")
reflinks = vim.bindeval('b:reflink_dict')
if not reflinks.has_key(id):
	reflinks[id] = '[{0}]: {1} "{2}"'.format(id, url, title)
snip.rv = id
`]$0
endsnippet
```

The idea is that you type a trigger like `rlink #id http://example.com The link title<tab>`
and get the trigger replaced with a snippet for a Markdown link reference:
`[{LINK_TEXT}][ID]` at the same time as a suitable link definition is saved to a buffer local dict variable, to be later written out elsewhere in the document with the help of another snippet. The `LINK_TEXT` defaults to the visually selected text if any, or else to the link title, while the id defaults to the link text reformatted into a valid id. Any of the 'arguments' in the trigger can be the name of a Vim register like `@+`, in which case they will be expanded to the content of that register (this is very handy when you have copied an URL in the web browser! :-)  The `html_id()` function is just some `regex.sub()` calls which massage the argument into a valid HTML id string. The `reg_or()` function lives in a module in `~/.vim/pythonx/` and looks like this:

```python
def reg_or(text,default,filtr=None):
    """
    -   If ``text`` looks like a Vim register name in ``@x`` notation 
            replace it with the content of that register.
    -   If ``text`` or the register is empty return ``default``.
    -   If ``filtr`` is a callable filter ``text`` through it, otherwise return
            ``text`` as is.

    This is useful when ``text`` is the content of a match group, avoiding to
    litter lots of snippets with the checking/replacing/filtering code.
    """
    if text is None:
        text = default
    if reg_name_re.match(text):
        text = vim.eval(text)
    if not len(text):
        text = default
    if callable(filtr):
        return filtr(text)
    else:
        return text
```

This works perfectly most of the time, but sometimes (especially, perhaps only, when there was a visual selection) I get junk characters before the expanded text like `v3G15|o3G2|o[the foo command][the_foo_command]`. On the face of it those characters are entirely unrelated to anything in the input -- come to think of it they look like normal mode commands. I don't know if this is because of something stupid I'm doing, a bug in UltiSnips or something in between, but I thought I should bring it to your attention at least to hopefully get an explanation what's going on. And in case you wonder, this madness saves lots of time, and keystrokes, for me when converting vanilla plaintext into Markdown. :-)

SirVer:
Fix junk characters appearing in snippet expansions (#751)

Issue #751 had two confirmed mechanisms that both produced the same kind of leaked move-command bytes in the buffer:

1. **`post_jump` action calls a typeahead-consuming Vim function** (e.g. `vim.command('call input("pause")')`). `input()` swallows the leading `<Esc>` of the queued move command and the rest (`a`, or `v3G10|o3G4|o<C-g>` for visual selections) is then processed in the original insert mode and typed as literal text. This is Maelan's minimal repro (2018).

2. **`post_jump` action calls `snip.expand_anon(...)` whose first tabstop is defaulted.** The outer jump's mode-entry uses `:startinsert`, which is scheduled — only effective after the current `:call` returns. The inner jump's `select()` then queues `v…|o…|o<C-g>` for the visual+select transition. When `:call` ends, `:startinsert` flips the mode to insert *before* vim processes the typeahead, so the visual-mode keys land in insert mode and end up as literal buffer text. This is the shape of nihlaeth's 2017 report.

bpj's original 2016 report (regex-trigger snippet with `vim.bindeval`) was not reproducible on master even before this fix — likely incidentally addressed by later changes.

Changes:

- `vim_helper.select()` now positions the cursor synchronously and *returns* the move command instead of feeding it. Two fast paths are added: (a) empty tabstop in insert mode — only the undo break is needed; (b) leaving select/visual mode is done synchronously via `\<C-\>\<C-N>` so a queued `<Esc>` can no longer be swallowed.
- `_jump()` is responsible for feeding the move command. It runs `do_post_jump` inside `_action_context()` so a nested `snip.expand_anon` is recorded in `_snip_expanded_in_action`, and skips feeding the outer move command when the flag is set — the inner jump already queued the correct mode-entry keys.

Regression tests in `test/test_PostJumpJunk.py` cover both mechanisms.

Fixes #751.


SirVer:
The test failures are unrelated; likely we rely on ordering of a set() somewhere. I'll investigate in a followup.

Issue #1377, opened 2021-08-05
Anoying error messages on every key pressed.


ajr-dev:
When I'm modifying my configuration, I keep receiving an error saying that it doesn't find Python. But if I keep receiving an error every time I press a key and it doesn't let me write I can't fix anything. I had to remove it from my configuration to be able to write.

SirVer:
The report does not include a vimrc, Vim/UltiSnips versions, or the actual error text, so I cannot reproduce it.

From the description ("doesn't find Python … error every time a key is pressed") this sounds like the same Python-not-loadable failure tracked in #1237: every `InsertLeave` / `CursorMovedI` autocmd runs `py3` which raises, and the error is re-surfaced on every keystroke. UltiSnips needs `+python3` and a working `python3_host_prog` (Neovim).

Quick checks:
- `:echo has('python3')` should return `1`.
- For Neovim, `:checkhealth provider.python` should show a working interpreter, or `:let g:python3_host_prog = '/path/to/python3'` in `init.vim`.

If you can post the exact error message and the output of `:version` / `:checkhealth`, I'm happy to reopen. Closing for now as unreproducible.

Issue #1209, opened 2020-04-17
Error Detected while processing 


Raoul555:
**Expected behavior:**

Open a text file in Vim and entering insert mode allows me to type in characters normally.

**Actual behavior:**

Since installing UltiSnips, every keystroke entered in insert mode results in the following error text:

```
Error detected while processing function UltiSnips#TrackChange:
line    1:
Traceback (most recent call last):
Press ENTER or type command to continue
```

An additional ENTER prompts the following:

```Error detected while processing function UltiSnips#TrackChange:
line    1:
  File "<string>", line 1, in <module>
Press ENTER or type command to continue
Error detected while processing function UltiSnips#TrackChange:
line    1:
NameError: name 'UltiSnips_Manager' is not defined
Press ENTER or type command to continue
Error detected while processing function UltiSnips#TrackChange:
line    1:
Traceback (most recent call last):
Press ENTER or type command to continue
Error detected while processing function UltiSnips#TrackChange:
line    1:
  File "<string>", line 1, in <module>
Press ENTER or type command to continue
```

**Steps to reproduce**

Invoke vim via a `git commit --amend`
Switch to edit mode, and type any text.
At first keystroke press while in editing mode, the error appears.

-----

- **OS**:

Debian 10

- **Vim Version**:

IM - Vi IMproved 8.1 (2018 May 18, compiled Jun 15 2019 16:41:15)
Included patches: 1-875, 878, 884, 948, 1046, 1365-1368, 1382, 1401
Modified by team+vim@tracker.debian.org
Compiled by team+vim@tracker.debian.org
Huge version with GTK2 GUI.  Features included (+) or not (-):
+acl               +digraphs          +lispindent        +persistent_undo   +textobjects
+arabic            +dnd               +listcmds          +postscript        +textprop
+autocmd           -ebcdic            +localmap          +printer           +timers
+autochdir         +emacs_tags        +lua               +profile           +title
-autoservername    +eval              +menu              -python            +toolbar
+balloon_eval      +ex_extra          +mksession         +python3           +user_commands
+balloon_eval_term +extra_search      +modify_fname      +quickfix          +vartabs
+browse            +farsi             +mouse             +reltime           +vertsplit
++builtin_terms    +file_in_path      +mouseshape        +rightleft         +virtualedit
+byte_offset       +find_in_path      +mouse_dec         +ruby              +visual
+channel           +float             +mouse_gpm         +scrollbind        +visualextra
[...]

PR #1658, opened 2026-05-14
Bail out cleanly when Python 3 is unavailable

### LINKED ISSUE #1237: Silent: Remove verbose error if python not loadable
Hi, awesome plugin, I `:packadd` it in my vimrc.

__Problem:__ But on many machines, vim python is not well instaled (my bad).
__Solution:__ It would be great if Ultisnip just silently fails in that case.
__My problem, my job:__ I can do a PR, now I'm asking if dev agrees with this __silent fail__.

Thanks guys.

PS : the message:
```text
Error detected while processing /home/almamgr/.vim/pack/bundle/opt/ultisnips/autoload/UltiS
nips.vim:
line    7:
E370: Could not load library /usr/lib/python3.8/config-3.8-x86_64-linux-gnu/libpython3.8.so

Error detected while processing /home/almamgr/.vim/pack/bundle/opt/ultisnips/autoload/UltiS
nips.vim:
line    7:
E263: Sorry, this command is disabled, the Python library could not be loaded.
Error detected while processing /home/almamgr/.vim/pack/bundle/opt/ultisnips/autoload/UltiS
nips.vim:
line    8:
E370: Could not load library /usr/lib/python3.8/config-3.8-x86_64-linux-gnu/libpython3.8.so

Error detected while processing /home/almamgr/.vim/pack/bundle/opt/ultisnips/autoload/UltiS
nips.vim:
line    8:
E263: Sorry, this command is disabled, the Python library could not be loaded.
Error detected while processing function UltiSnips#TrackChange:
line    1:
E370: Could not load library /usr/lib/python3.8/config-3.8-x86_64-linux-gnu/libpython3.8.so

Error detected while processing function UltiSnips#TrackChange:
line    1:
E263: Sorry, this command is disabled, the Python library could not be loaded.
```

SirVer:
Vim happily kept calling `UltiSnips#TrackChange` from `TextChangedI` / `InsertCharPre` on every keystroke even when `py3` could not load libpython (E370 / E263) or when the UltiSnips package itself was not on `sys.path` (NameError on `UltiSnips_Manager`). Each keystroke re-spammed the same stack trace, making Vim unusable on hosts where the plugin is checked into a shared dotfiles repo but Python is missing.

At plugin-load time, force the lazy `py3` import now and ask Python to flip a Vim-side flag. If either the libpython load or `from UltiSnips import UltiSnips_Manager` fails, emit one `WarningMsg` message (suppressible via the documented `g:UltiSnipsNoPythonWarning`, see `:help UltiSnips-python-warning`) and `finish` before defining any commands, autocmds, or mappings.

The opt-out variable was already documented but never wired up; this patch finally implements the behaviour the docs promise.

Closes #1237. Also addresses the symptom reported in #1209 and #1377 when the import fails for any other reason.

### DIFF
--- a/plugin/UltiSnips.vim
+++ b/plugin/UltiSnips.vim
@@ -13,29 +13,32 @@ if !has('nvim') && version < 820
    finish
 endif
 
+" UltiSnips needs Python 3. Bail with a clear message if Vim wasn't
+" compiled with python3 support at all, so we don't fall through to the
+" runtime-check below (which would just say "Python is broken" without
+" distinguishing missing-from-build vs broken-at-load).
+if !has('python3') && !get(g:, 'UltiSnipsNoPythonWarning', 0)
+    echohl WarningMsg
+    echom 'UltiSnips requires Vim compiled with +python3 support; disabling UltiSnips.'
+    echom '          See :help UltiSnips-requirements for setup.'
+    echom '          Silence this message with `let g:UltiSnipsNoPythonWarning = 1`.'
+    echohl None
+endif
+if !has('python3')
+    finish
+endif
 
-" TODO(robot): should we not have here a simple block to check for python3,
-" similar to the warning above for the required vim version? This means, this
-" check we no longor need.
-
-" Bail out cleanly when Python 3 is unusable, instead of letting every
-" autocmd and mapping below re-raise E370/E263 (Vim can't load libpython)
-" or `NameError: UltiSnips_Manager` (the package isn't on the Python path)
-" on every keystroke (#1237, #1209).
-"
 " `has('python3')` is necessary but not sufficient: Vim can be compiled
-" with dynamic Python support and still fail to load the library at
-" runtime. We force the lazy load now by asking Python to flip a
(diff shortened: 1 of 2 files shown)


plugin/UltiSnips.vim
```diff
@@ -13,6 +13,60 @@ if !has('nvim') && version < 820
...
+try
+    py3 << EOF
+try:
+    import vim
+    from UltiSnips import UltiSnips_Manager
+    vim.command('let s:ultisnips_python3_ok = 1')
+except Exception as _err:
+    import traceback as _tb
+    _msg = _tb.format_exception_only(type(_err), _err)[-1].strip()
+    vim.command("let s:ultisnips_python3_error = " + repr(_msg))
```
wookayin, line 49: Can we have an option to emit a full stacktrace? It hides debugging information on some errors which makes it difficult for me to troubleshoot the issue.

Maybe something like healthcheck (for neovim only, though)?

SirVer: we figured it out in the end.

PR #1664, opened 2026-05-14
Add UltiSnips#FindAutoTriggerConflicts() for #1506

### LINKED ISSUE #1506: Detecting conflicting snippet triggers
Hello,

Is there a way to detect incompatible snippet triggers?

Say:

```
snippet st "superscript" iA
<---snippet body--->
endsnippet
```

vs

```
snippet bst "binary spanning tree data" iA
<---snippet body--->
endsnippet
```

vs

```
snippet bs "basis information" iA
<---snippet body--->
endsnippet
```
----

In the above case, `bs` would fire when `bst` is typed, so from the plugin POV there is no conflict, per se.

But from the user POV, `bst` will never fire. Is there a way to possibly call a function in UltiSnips that will detect and report such cases?

Thanks.

SirVer:
When a snippet uses the `A` (autotrigger) flag and its trigger is a literal prefix of another snippet's trigger, the autotrigger fires the moment the user finishes typing the short trigger - so the longer trigger can never be reached. The bug-report author had this exact pain with their TeX snippet collection (`bs` shadowing `bst`, etc.) and ended up writing an external C++ utility to flag conflicts.

This PR provides it as a built-in. `UltiSnips#FindAutoTriggerConflicts()` walks the snippets visible in the current scope (including those inherited via `extends`), picks the `A`-flagged simple triggers, and returns the `[short, long]` pairs where the autotrigger is a prefix of another trigger. Regex triggers (`r`) are skipped because prefix containment isn't well-defined for them. Documented next to the existing predicate trio, with the original failing example as the doc.

Alternatives considered:

1. **This PR**: dedicated entry point that returns the pairs. Cheapest for users (one call, structured result), easy to wire into a `:checkhealth`-style command or CI.
2. **Extend `UltiSnips#SnippetsInCurrentScope`**: add `options` to `ulti_dict_info` and let users compute conflicts themselves. More flexible but pushes the prefix-detection algorithm onto every caller, and the existing function's contract is "describe what could expand here", not "describe everything for analytics".
3. **Wontfix and point to external tooling**: keeps the API tight but ignores a real workflow concern that the author already hit hard enough to write C++ for.

Going with option 1.

Fixes #1506.

### DIFF
--- a/autoload/UltiSnips.vim
+++ b/autoload/UltiSnips.vim
@@ -126,6 +126,10 @@ function! UltiSnips#ToggleAutoTrigger() abort
     return py3eval("UltiSnips_Manager._toggle_autotrigger()")
 endfunction
 
+function! UltiSnips#FindAutoTriggerConflicts() abort
+    return py3eval("UltiSnips_Manager.find_autotrigger_prefix_conflicts()")
+endfunction
+
 function! UltiSnips#SaveLastVisualSelection() range abort
     py3 UltiSnips_Manager._save_last_visual_selection()
     return ""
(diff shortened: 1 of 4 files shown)


SirVer:
I feel this code is adding a lot of complexity for a relatively small pain point. I decided to not merge this.

PR #1683, opened 2026-05-14
Source defaults before Python import; gate after-plugin remap

### LINKED ISSUE #1679: Regression: errors on startup if trigger key mappings are not set
TL;DR (update):

If any of the following variables is missing in user config, Ultisnips will error on startup.

```
g:UltiSnipsExpandTrigger
g:UltiSnipsJumpForwardTrigger
g:UltiSnipsJumpBackwardTrigger
```

----

Original message:

#1658 (in v4.0) probably has introduced a regression: the warning message happens at all times for me.

```
UltiSnips: Python 3 is present but unusable; disabling UltiSnips.
           KeyError: 'Key not found: UltiSnipsExpandTrigger'
           See :help UltiSnips-requirements for setup.
           Silence this message with `let g:UltiSnipsNoPythonWarning = 1`.
```

I'll try to make a exact minimal repro, but wanted to flag the error asap before you @SirVer cut a release. ~~FYI I'm lazy-loading python3 provider nvim, so it might be possible that this is due to a timing/ordering issue.~~ Update: it has nothing to do with ordering/lazy loading.

As for the error, I don't have `g:UltiSnipsExpandTrigger` config in my vimrc/init.lua; it's a bit tangential #1605 might be the cause, and probably the error was not visible until #1658 landed.

SirVer:
Source defaults before Python import; gate after-plugin remap

`plugin/UltiSnips.vim` forces the `py3 from UltiSnips import UltiSnips_Manager` at plugin load (added in #1658, so a libpython load failure or missing-package error is caught with one warning instead of being re-raised on every keystroke). The Python module reads `g:UltiSnipsExpandTrigger`, `g:UltiSnipsJumpForwardTrigger`, and `g:UltiSnipsJumpBackwardTrigger` at import time via `vim.vars[...]` (the form without a fallback, switched to in #1605). The defaults for those globals are set as a side effect of `autoload/UltiSnips/map_keys.vim` being sourced, which only happened from the `call UltiSnips#map_keys#MapKeys()` line at the *end* of `plugin/UltiSnips.vim` — i.e. after the Python import. A user who hadn't set the triggers in their vimrc therefore saw:

```
UltiSnips: Python 3 is present but unusable; disabling UltiSnips.
           KeyError: 'Key not found: UltiSnipsExpandTrigger'
```

even with a perfectly working Python 3.

`plugin/UltiSnips.vim` now runs `runtime autoload/UltiSnips/map_keys.vim` immediately after the `has('python3')` build check and *before* the Python try block. Sourcing the autoload runs its `if !exists | let g:... = ... | endif` defaults block but does not call the function, so the public globals are defaulted in time for the Python module to read them at import. The `call UltiSnips#map_keys#MapKeys()` line stays at the bottom of the file so the keymaps are only installed once the Python check has passed — a load failure can `finish` cleanly without leaving the plugin half-initialised.

`after/plugin/UltiSnips_after.vim` (the "reclaim keys for SuperTab" hook) is sourced unconditionally by Vim regardless of whether `plugin/UltiSnips.vim` finished early; gate it on `exists(':UltiSnipsEdit')` (the command is only registered after the Python check passes) so it doesn't reinstall the keymaps on top of a half-loaded plugin either.

Fixes #1679.

### DIFF
--- a/plugin/UltiSnips.vim
+++ b/plugin/UltiSnips.vim
@@ -13,6 +13,46 @@ if !has('nvim') && version < 820
    finish
 endif
 
+" Default the public globals before the Python import below. The autoload
+" file `autoload/UltiSnips/map_keys.vim` used to default these as a
+" side-effect of being sourced from `MapKeys()`, which ran *after* the
+" Python import — and the Python module reads
+" `g:UltiSnipsExpandTrigger` / `g:UltiSnipsJumpForwardTrigger` /
+" `g:UltiSnipsJumpBackwardTrigger` at module load, so a user who hadn't
+" set them in their vimrc would see a spurious "Python 3 is present but
+" unusable" warning with a `KeyError: 'Key not found:
+" UltiSnipsExpandTrigger'`. See #1679.
+"
+" NOTE: expansion and forward jumping can, but needn't be, the same
+" trigger.
+if !exists('g:UltiSnipsExpandTrigger')
+    let g:UltiSnipsExpandTrigger = '<tab>'
+endif
+if !exists('g:UltiSnipsListSnippets')
+    let g:UltiSnipsListSnippets = '<c-tab>'
+endif
+if !exists('g:UltiSnipsJumpForwardTrigger')
+    let g:UltiSnipsJumpForwardTrigger = '<c-j>'
+endif
+if !exists('g:UltiSnipsJumpBackwardTrigger')
+    let g:UltiSnipsJumpBackwardTrigger = '<c-k>'
+endif
+if !exists('g:UltiSnipsRemoveSelectModeMappings')
+    let g:UltiSnipsRemoveSelectModeMappings = 1
+endif
+if !exists('g:UltiSnipsMappingsToIgnore')
+    let g:UltiSnipsMappingsToIgnore = []
+endif
+if !exists('g:UltiSnipsEditSplit')
+    let g:UltiSnipsEditSplit = 'normal'
+endif
+if !exists('g:UltiSnipsSnippetDirectories')
+    let g:UltiSnipsSnippetDirectories = ['UltiSnips']
+endif
+if !exists('g:UltiSnipsEnableSnipMate')
+    let g:UltiSnipsEnableSnipMate = 1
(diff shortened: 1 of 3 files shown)


plugin/UltiSnips.vim
```diff
@@ -13,6 +13,46 @@ if !has('nvim') && version < 820
...
+" `g:UltiSnipsJumpBackwardTrigger` at module load, so a user who hadn't
+" set them in their vimrc would see a spurious "Python 3 is present but
+" unusable" warning with a `KeyError: 'Key not found:
+" UltiSnipsExpandTrigger'`. See #1679.
+"
+" NOTE: expansion and forward jumping can, but needn't be, the same
+" trigger.
+if !exists('g:UltiSnipsExpandTrigger')
+    let g:UltiSnipsExpandTrigger = '<tab>'
+endif
```
wookayin: In my config, I explicitly turned this off so that Ultisnips does not *override* an existing keymap `<tab>`.

See `autoload/UltiSnips/map_keys.vim` -- 
`UltiSnips#map_keys#MapKeys()` would not create a keymap if the variable is not set.

There are also duplicate codes there that would set the default value of the global variable (which can be removed or consolidated).

So I'd suggest we should instead keep `g:UltiSnipsExpandOrJumpTrigger` optional, and `UltiSnips_Manager` accept these three triggers as optional.

SirVer: @wookayin What do you mean with "explicitly turn these off"?

the functionality should not be changed - the variables was always set, but a bit later in map_keys.vim (which is implicitly sourced by the call to `UltiSnips#MapKeys` later in this file. 

> There are also duplicate codes there that would set the default value of the global variable (which can be removed or consolidated).

Again, I do not understand. Can you clarify?

> So I'd suggest we should instead keep g:UltiSnipsExpandOrJumpTrigger optional, and UltiSnips_Manager accept these three triggers as optional.

It is optional - as it was before. Am I missing something?

SirVer: clarified in #1679.

PR #1687, opened 2026-05-18
Correct VIM version check

claw0ry:
Fixes the VIM 8.2 version check on startup to check against the correct value.


SirVer:
Ups, nice catch. thanks!

